Issue #4, opened 2019-07-15
Support Polymorphic serialization


frzme:
It would be nice if Polymorphic (de)serialization was supported in the same way it is for JSON.

(see https://medium.com/@andreclassen1337/goodbye-runtimetypeadapterfactory-polymorphic-serialization-using-kotlinx-serialization-46a8cec36fdc )

charleskorn:
Hi @frzme, thanks for the suggestion!

I don't have capacity to add this in the short term, so if you're keen, I'm happy to provide some pointers and review a PR as you go. I'm imagining this would use YAML's local tags syntax (`!square` / `!circle`) to indicate the type - https://camel.readthedocs.io/en/latest/yamlref.html#tags has a good explanation of this. Was that what you had in mind, or were you thinking of something else?

frzme:
Hi @charleskorn I was not aware of this feature and thus was thinking of basically the same as in json - but if it yaml has native support for it it sounds great to use it.

Unfortunatly I most likely also won't have capacity to work on it anytime soon.

charleskorn:
This has been resolved with #7 and will be part of the next release.

PR #7, opened 2019-09-09
Support tagged nodes

EdwarDDay:
- implement #4 
- support yaml tags for polymorphic de-/serialization

### DIFF
--- a/src/main/kotlin/com/charleskorn/kaml/YamlOutput.kt
+++ b/src/main/kotlin/com/charleskorn/kaml/YamlOutput.kt
@@ -44,19 +44,12 @@ import java.util.Optional
 
 internal class YamlOutput(
     writer: StreamDataWriter,
-    private val yaml: Yaml
+    override val context: SerialModule,
+    private val configuration: YamlConfiguration
 ) : ElementValueEncoder() {
     private val settings = DumpSettingsBuilder().build()
     private val emitter = Emitter(settings, writer)
     private var currentTag: String? = null
-    private val configuration: YamlConfiguration get() = yaml.configuration
-    override val context: SerialModule get() = yaml.context
-    private val yamlTag: Optional<String>
-        get() {
-            return Optional.ofNullable(currentTag).also {
-                currentTag = null
-            }
-        }
 
     init {
         emitter.emit(StreamStartEvent())
@@ -100,7 +93,7 @@ internal class YamlOutput(
     }
 
     override fun beginStructure(desc: SerialDescriptor, vararg typeParams: KSerializer<*>): CompositeEncoder {
-        val tag = yamlTag
+        val tag = getAndClearTag()
         val implicit = !tag.isPresent
         when (desc.kind) {
             is StructureKind.LIST -> emitter.emit(SequenceStartEvent(Optional.empty(), tag, implicit, FlowStyle.BLOCK))
@@ -120,7 +113,13 @@ internal class YamlOutput(
     }
 
     private fun emitScalar(value: String, style: ScalarStyle) =
-        emitter.emit(ScalarEvent(Optional.empty(), yamlTag, ALL_IMPLICIT, value, style))
+        emitter.emit(ScalarEvent(Optional.empty(), getAndClearTag(), ALL_IMPLICIT, value, style))
+
+    private fun getAndClearTag(): Optional<String> {
+        val tag = Optional.ofNullable(currentTag)
+        currentTag = null
+        return tag
+    }
 
     companion object {
         private val ALL_IMPLICIT = ImplicitTuple(true, true)
--- a/src/main/kotlin/com/charleskorn/kaml/YamlOutput.kt
+++ b/src/main/kotlin/com/charleskorn/kaml/YamlOutput.kt
@@ -100,8 +100,8 @@ internal class YamlOutput(
     }
 
     override fun beginStructure(desc: SerialDescriptor, vararg typeParams: KSerializer<*>): CompositeEncoder {
-        val tag: Optional<String> = yamlTag
-        val implicit = tag.isEmpty
+        val tag = yamlTag
+        val implicit = !tag.isPresent
         when (desc.kind) {
             is StructureKind.LIST -> emitter.emit(SequenceStartEvent(Optional.empty(), tag, implicit, FlowStyle.BLOCK))
             is StructureKind.MAP, StructureKind.CLASS -> emitter.emit(MappingStartEvent(Optional.empty(), tag, implicit, FlowStyle.BLOCK))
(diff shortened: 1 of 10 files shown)


charleskorn:
Hey @EdwarDDay, thanks for the PR! Before I take a closer look, could you please fix the compilation error Travis is reporting?

src/main/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -41,10 +44,19 @@ import java.util.Optional
 
 internal class YamlOutput(
     writer: StreamDataWriter,
-    private val configuration: YamlConfiguration
+    private val yaml: Yaml
```
charleskorn: `YamlInput` takes the configuration and context as properties, rather than receiving the whole `Yaml` instance - it would be good to keep it consistent with that

src/main/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -41,10 +44,19 @@ import java.util.Optional
...
 ) : ElementValueEncoder() {
     private val settings = DumpSettingsBuilder().build()
     private val emitter = Emitter(settings, writer)
+    private var currentTag: String? = null
+    private val configuration: YamlConfiguration get() = yaml.configuration
+    override val context: SerialModule get() = yaml.context
+    private val yamlTag: Optional<String>
```
charleskorn: Is there a reason why this property mutates the state? Maybe instead of this, in `beginStructure`, we can construct the `Optional.ofNullable(...)` value and then set `currentTag` to `null`? Or maybe turn this property into a method that does what this getter is doing?

src/test/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -831,6 +838,119 @@ object YamlReadingTest : Spek({
...
+                                SealedWrapper(TestSealedStructure.SimpleSealedInt(-987)),
+                                SealedWrapper(TestSealedStructure.SimpleSealedInt(654)),
+                                SealedWrapper(TestSealedStructure.SimpleSealedString("tests"))
+                            )
+                        )
+                    }
+                }
+            }
+
+            val interfaceYaml = Yaml(context = interfaceModule)
```
charleskorn: Is there something special that we have to do with interfaces as compared to classes? I'm not clear if these tests are required.

src/test/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -857,7 +977,55 @@ object YamlReadingTest : Spek({
...
+                            line { toBe(1) }
+                            column { toBe(1) }
+                        }
+                    }
+                }
+            }
+
+            context("given some input representing an object with a tagged value as a key") {
+                val input = """
+                    !<sealedInt> { }: something
```
charleskorn: To make the intent of this test clear, could you remove the `{ }` and replace it with a string (eg. `!<sealedInt> thing: something`)?

src/test/kotlin/com/charleskorn/kaml/YamlTaggedTest.kt
```diff
@@ -0,0 +1,105 @@
...
+
+import ch.tutteli.atrium.api.cc.en_GB.notToThrow
+import ch.tutteli.atrium.api.cc.en_GB.toBe
+import ch.tutteli.atrium.verbs.assert
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
+object YamlTaggedTest : Spek({
+    describe("a YAML tagged node") {
+        describe("creating an instance") {
```
charleskorn: This kind of test is only necessary when we do some kind of validation in the constructor (eg. line we do in `YamlMap`)

src/test/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -831,6 +838,119 @@ object YamlReadingTest : Spek({
                 }
             }
 
+            val sealedYaml = Yaml(context = sealedModule)
+
+            context("given some tagged input representing an object where the resulting type should be a sealed class (int)") {
+                val input = """
+                    element: !<sealedInt>
```
charleskorn: Is there any way to support `!sealedInt` instead of `!<sealedInt>`?

src/test/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
```diff
@@ -471,6 +478,118 @@ object YamlWritingTest : Spek({
...
+                            value: -20
+                        - element: !<sealedString>
+                            value: "another test"
+                    """.trimIndent()
+                    )
+                }
+            }
+        }
+
+        describe("handling interface") {
```
charleskorn: Again, is there anything special we have to do for interfaces? Otherwise I would remove this set of extra tests

src/test/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
EdwarDDay: That's not yet possible, but I can create another PR after that one. Currently scalar nodes with tags are not supported and supporting them would be a new feature with a new PR in my opinion.

src/test/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
EdwarDDay: I think that can be done via the `%TAG` directive, but I'm not sure how to do that.

charleskorn:
Hey @EdwarDDay, thanks for incorporating my feedback and apologies for the delay in getting back to you. This is looking really good. The only things I'd like to see before I merge this are:

* More tests for reading tagged input nodes - for example, the logic in `YamlTaggedInput.getCurrentLocation()` is never exercised (take a look at the tests for other kinds of nodes for how this is used to generate exceptions with location information) and none of the scalar decoding methods (eg. `decodeInt()`) are exercised. I'm not suggesting to test these scalar decoding methods individually, but it does seem that they're not being used by the current tests which suggests they might not be necessary and can be removed.
* If the scalar decoding methods are required, I'd like to see them simplified a bit to remove the duplication and risk of things getting out of sync - for example, I'm imagining `decodeInt()` could look something like:

    ```kotlin
    override fun decodeInt(): Int = decodeScalar { decodeInt() }
    ```

   where `decodeScalar` is a private inline function that does the check for the current index and picks the appropriate decoder

EdwarDDay:
I removed the `decode[Int|Short|Unit|...]` methods, because these are not yet used. The `getCurrentLocation()` is now tested. Additionally a tag is ignored, if it is not needed.

charleskorn:
Looks good to me, thanks @EdwarDDay!

PR #9, opened 2019-09-20
Avoid a duplicated constant

sksamuel:


charleskorn:
Looks good to me, thanks @sksamuel!

PR #10, opened 2019-09-23
Support tagged scalars

EdwarDDay:


src/main/kotlin/com/charleskorn/kaml/YamlParser.kt
```diff
@@ -60,23 +59,12 @@ class YamlParser(yamlSource: String) {
...
+            return retrieve()
         } catch (e: MarkedYamlEngineException) {
             throw translateYamlEngineException(e)
         }
     }
 
-    private fun checkForUnsupportedFeatures(event: Event) {
-        if (event is ScalarEvent) {
-            if (event.tag.isPresent) {
-                throw UnsupportedYamlFeatureException("tags", event)
```
charleskorn: Given that `UnsupportedYamlFeatureException` won't be thrown any more, could you please remove the class?

src/test/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -945,6 +960,65 @@ object YamlReadingTest : Spek({
...
 
+            val simpleYaml = Yaml(context = simpleModule)
+
+            context("given some simple int input representing an object") {
+                val input = """
+                    test: !<simpleInt> 42
+                """.trimIndent()
+
+                context("parsing that input") {
+                    SimpleWrapper.serializer().list
```
charleskorn: This doesn't need to be here :)

src/test/kotlin/com/charleskorn/kaml/testobjects/TestObjects.kt
```diff
@@ -63,3 +79,126 @@ val sealedModule = SerializersModule {
         TestSealedStructure.SimpleSealedString::class with TestSealedStructure.SimpleSealedString.serializer()
     }
 }
+
```
charleskorn: This file is getting quite large - how about we split the polymorphic serialization stuff out to its own file in this folder?

src/test/kotlin/com/charleskorn/kaml/testobjects/TestObjects.kt
```diff
@@ -63,3 +79,126 @@ val sealedModule = SerializersModule {
...
+interface SimpleInterface
+
+class SimpleNull : SimpleInterface {
+    companion object {
+        val kSerializer: KSerializer<SimpleNull> = UnitSerializer.nullable.mapped("simpleNull", { SimpleNull() }, { null })
+    }
+
+    override fun equals(other: Any?): Boolean = other is SimpleNull
```
charleskorn: Are `equals` and `hashCode` required here? Could we use an `object` instead of a `class` perhaps?

src/main/kotlin/com/charleskorn/kaml/YamlInput.kt
```diff
@@ -334,7 +347,7 @@ private class YamlTaggedInput(val taggedNode: YamlTaggedNode, context: SerialMod
         isPolymorphic = kind === UnionKind.POLYMORPHIC
     }
 
-    private inline fun <T> maybeCallOnChild(blockOnTag: () -> T, blockOnChild: YamlInput.() -> T): T {
+    private inline fun <T> maybeCallOnChild(functionName: String = "", blockOnTag: () -> T = { throw IllegalArgumentException("can't call $functionName on tag") }, blockOnChild: YamlInput.() -> T): T {
```
charleskorn: Rather than the default values for parameters, could this be rewritten as two overloads? I'm imagining one that takes the two lambdas, and another that takes the function name and `blockOnChild`

charleskorn:
Looks good to me, thanks @EdwarDDay!

PR #14, opened 2020-03-06
Migrate to kotlin 1.3.70 and kotlinx-serialization 0.20.0

HeartPattern:
- Use {type}.serializer() method instead of {type}Serializer class
 - Use list and nullable extension property in builtin package
 - Use MapSerializer class instead of map extension property
 - Use object to serialize kotlin.Unit
 - Use AbstractEncoder/Decoder instead of ElementValueEncoder/Decoder
 - Implement decodeElementIndex in AbstractDecoder
 - Change parameter name of implemented method to match with super method

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -31,12 +31,12 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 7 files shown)


charleskorn:
Thanks for the PR @HeartPattern!

PR #15, opened 2020-03-06
Remove dependency to internal serialization API

EdwarDDay:


charleskorn:
Thanks for the PR @EdwarDDay!

Issue #33, opened 2020-08-24
Please provide output format settings


Vampire:
It would be nice if it were possible to be able to influence output format settings like indentation depth, width for line-breaks and similar. Not everyone is happy with SnakeYamls defaults. :-)

charleskorn:
Hey @Vampire, thanks for the feedback. Do these correspond with snakeyaml's `setIndent` and `setWidth` methods on [`DumpSettingsBuilder`](https://bitbucket.org/asomov/snakeyaml-engine/src/master/src/main/java/org/snakeyaml/engine/v2/api/DumpSettingsBuilder.java), or are you after something else?

Vampire:
I guess so, I'm not familiar with SnakeYaml, but I'd like to be able to configure everything format-related the underlying emitter supports.
I'm for example using kaml to read in a GitHub workflow file that has many anchors in it which is not supported by GitHub yet.
But I would need many repetition without anchors and I don't like that.
So I used anchors and now use kaml to read in the file and thereby resolve the anchors and then write out the file again.
So I'm not producing some protocol payload but a file that is checked into version control.
And with the default settings, this file looks not too nice.
By being able to change the output options I hope being able to produce a bit better readable result.

charleskorn:
@Vampire I've added support for both of those settings, and I'll release a new version soon. Please take a look and let me know if there are further options you would like to see.

knightzmc:
To build on this, it would be nice if we could customise the output style of sequences (between block style and flow style).
The system I'm working on expects them in flow style, so unit tests fail (since they literally compare the yaml contents)

I appreciate this isn't a fault of Kaml, but it would be a nice option to have nonetheless

charleskorn:
Given there's been no activity on this issue for a while and two PRs have been merged for it, I'm going to go ahead and close this issue. If there are other formatting options you'd like to see, please open a new issue.

PR #40, opened 2020-08-30
Add support for different flow styles in Sequences

knightzmc:
This adds support for the Flow Style of rendering Sequences, by adding a parameter to `YamlConfiguration`

The default block style is still used to avoid breaking anything.

This is referenced in #33

### DIFF
--- a/src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -31,6 +31,7 @@ import org.snakeyaml.engine.v2.common.FlowStyle
  *    * [PolymorphismStyle.Property]: use a property (eg. `{ type: typeOfThing, property: value }`)
  * * [encodingIndentationSize]: number of spaces to use as indentation when encoding objects as YAML
  * * [breakScalarsAt]: maximum length of scalars when encoding objects as YAML (scalars exceeding this length will be split into multiple lines)
+ * * [sequenceStyle]: how sequences (aka lists and arrays) should be formatted. See [SequenceStyle] for an example of each
  */
 public data class YamlConfiguration constructor(
     internal val encodeDefaults: Boolean = true,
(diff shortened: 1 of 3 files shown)


charleskorn:
Hey @knightzmc, thanks for the PR. It looks like the formatting check is failing - could you please run `./gradlew spotlessApply` to fix any issues and push that change?

knightzmc:
Oh, that's weird. When I ran the tests it told me that spotless had failed so I formatted it. Must've accidentally changed something

knightzmc:
@charleskorn I ran spotless again, it seemed to have fixed all of the formatting issues.

src/test/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
```diff
@@ -362,7 +478,8 @@ object YamlWritingTest : Spek({
                     )
                 )
 
-                val serializer = MapSerializer(String.serializer(), MapSerializer(String.serializer(), String.serializer()))
+                val serializer =
```
charleskorn, line 481: Could you please revert all of the unrelated formatting changes here and below? (this helps keep the diff clean and focused on the changes you've made)

src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -36,10 +38,16 @@ public data class YamlConfiguration constructor(
     internal val extensionDefinitionPrefix: String? = null,
     internal val polymorphismStyle: PolymorphismStyle = PolymorphismStyle.Tag,
     internal val encodingIndentationSize: Int = 2,
-    internal val breakScalarsAt: Int = 80
+    internal val breakScalarsAt: Int = 80,
+    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block
```
charleskorn, line 42: Could you please add a short description of this parameter in the documentation comment above?

charleskorn: Something like this would be enough:

```
[sequenceStyle]: how to format sequences (lists or arrays) when encoding them as YAML
```

charleskorn (changes requested):
Thanks again for this PR. It looks good to me - there are only two minor things and then I think this should be good to merge.

knightzmc:
Hi @charleskorn I've gone through the diff and reverted all of the accidental test changes, along with adding a simple doc

charleskorn:
Looks great, thanks again for the PR @knightzmc!

PR #47, opened 2020-09-16
Support custom polymorphic key

ChanTsune:
I'm sorry for the sudden PR.

For personal reasons, I created a polymorphic key that can be changed to something other than "type", such as the `classDiscriminator` option in` kotlinx.serialization.json.Json`.

If this proposal is accepted, I would be grateful if you could incorporate it into the main branch.

### DIFF
--- a/src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -38,6 +38,7 @@ public data class YamlConfiguration constructor(
     internal val strictMode: Boolean = true,
     internal val extensionDefinitionPrefix: String? = null,
     internal val polymorphismStyle: PolymorphismStyle = PolymorphismStyle.Tag,
+    internal val polymorphismPropertyName: String = "type",
     internal val encodingIndentationSize: Int = 2,
     internal val breakScalarsAt: Int = 80,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -38,6 +38,7 @@ public data class YamlConfiguration constructor(
     internal val strictMode: Boolean = true,
     internal val extensionDefinitionPrefix: String? = null,
     internal val polymorphismStyle: PolymorphismStyle = PolymorphismStyle.Tag,
+    internal val polymorphismPropertyName: String = "type",
```
charleskorn, line 41: Could you please add a short explanation in the documentation above? Something like:

```
* [polymorphismPropertyName]: property name to use when [polymorphismStyle] is [PolymorphismStyle.Property]
```

charleskorn (commented):
Looks great, thanks for the PR @ChanTsune. I only have one minor nitpick and then I'm happy to merge this.

src/main/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
ChanTsune: Thank you for accepting PR.

I added documentation comment.

I am happy to contribute to kaml project!

PR #83, opened 2021-01-28
More elegant way to call the library's API

akuleshov7:
### What's done:
1) Added wrappers to call kaml in more elegant way
2) Added tests and documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ val input = """
     """.trimIndent()
 
 val result = Yaml.default.decodeFromString(Team.serializer(), input)
+// or using more elegant way:
+// val result = Yaml.decodeFromString<Team>(input)
 
 println(result)
 ```
@@ -47,6 +49,8 @@ data class Team(
 val input = Team("Amy", listOf("Bob", "Cindy", "Dan"))
 
 val result = Yaml.default.encodeToString(Team.serializer(), input)
+// or using more elegant way:
+// val result = Yaml.encodeToString<Team>(input)
 
 println(result)
 ```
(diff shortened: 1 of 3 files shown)


akuleshov7:
@charleskorn hey, what do you think about that?

charleskorn:
Hi @akuleshov7, thanks for the PR.

I'm not sure including this is a good idea. The documentation for [`serializer()`](https://github.com/Kotlin/kotlinx.serialization/blob/ef53d75/core/commonMain/src/kotlinx/serialization/Serializers.kt#L106) warns against using it: 

> This method uses platform-specific reflection available for the given erased `KClass` and is not recommended to use this method for anything, but last-ditch resort

I'd prefer that anyone who really wants to use this shortcut have to add an extension function themselves and understand the implications of what they're doing.

akuleshov7:
@charleskorn but you suggest to use `serializer()` directly in the readme: 
```
val result = Yaml.default.decodeFromString(Team.serializer(), input)
```

There are no other ways to call kaml application without calling serializer()

charleskorn:
Hi @akuleshov7, there are multiple different implementations of `serializer()`.

When you use something like `Team.serializer()`, this is resolved at compile time and works on all platforms. The compiler will report an error if the type doesn't support serialization and compilation will fail.

When you use the `KClass` extension method `serializer()`, as this PR does, the serializer is resolved at runtime, which has all the limitations discussed in [the documentation for that extension method](https://github.com/Kotlin/kotlinx.serialization/blob/ef53d75/core/commonMain/src/kotlinx/serialization/Serializers.kt#L106). If the serializer can't be determined at runtime, an exception is thrown.

akuleshov7:
Got it, ok.

PR #127, opened 2021-05-02
Add `decodeFromStream`

### LINKED ISSUE #126: Support reading yaml from InputStream
Currently, what must be done is
```
Yaml().decodeFromString(inputStream.bufferedReader().readText())
```

I propose having an helper function that reads directly from a stream:

```
Yaml().decodeFromStream(inputStream) // Same result as above
```

LeoColman:
Closes #126

### DIFF
--- a/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
+++ b/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
@@ -22,16 +22,25 @@ import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.StringFormat
+import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.modules.EmptySerializersModule
 import kotlinx.serialization.modules.SerializersModule
 import org.snakeyaml.engine.v2.api.StreamDataWriter
+import java.io.InputStream
 import java.io.StringWriter
+import java.nio.charset.Charset
+import kotlin.text.Charsets.UTF_8
 
 @OptIn(ExperimentalSerializationApi::class)
 public actual class Yaml(
     override val serializersModule: SerializersModule = EmptySerializersModule,
     public actual val configuration: YamlConfiguration = YamlConfiguration()
 ) : StringFormat {
+
+    public inline fun <reified T> decodeFromStream(inputStream: InputStream, charset: Charset = UTF_8): T {
+        return decodeFromString(inputStream.bufferedReader(charset).readText())
+    }
+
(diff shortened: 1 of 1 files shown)


src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
```diff
@@ -22,16 +22,25 @@ import kotlinx.serialization.DeserializationStrategy
...
 
 @OptIn(ExperimentalSerializationApi::class)
 public actual class Yaml(
     override val serializersModule: SerializersModule = EmptySerializersModule,
     public actual val configuration: YamlConfiguration = YamlConfiguration()
 ) : StringFormat {
+
+    public inline fun <reified T> decodeFromStream(inputStream: InputStream, charset: Charset = UTF_8): T {
+        return decodeFromString(inputStream.bufferedReader(charset).readText())
+    }
```
LeoColman, line 42: This can be added as an extension function instead of a member function. IDK which is better.

charleskorn: I'm happy with it as a member function.

src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
```diff
@@ -22,16 +22,25 @@ import kotlinx.serialization.DeserializationStrategy
...
+import kotlin.text.Charsets.UTF_8
 
 @OptIn(ExperimentalSerializationApi::class)
 public actual class Yaml(
     override val serializersModule: SerializersModule = EmptySerializersModule,
     public actual val configuration: YamlConfiguration = YamlConfiguration()
 ) : StringFormat {
+
+    public inline fun <reified T> decodeFromStream(inputStream: InputStream, charset: Charset = UTF_8): T {
+        return decodeFromString(inputStream.bufferedReader(charset).readText())
```
charleskorn, line 41: It might be worth changing `YamlParser` to accept a `Reader`, and then pass a `StringReader` in `decodeFromString` and an `InputStreamReader` here in `decodeFromStream` - this avoids trying to read the entire stream into a string, which could be quite large. What do you think?

charleskorn (commented):
Let me know what you think about making this slightly larger change - I think this will make things a little more efficient.

src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
LeoColman: Sorry for the delay.
If we're decoding to a Yaml object, it will always have to read the entire thing at once, won't it?

charleskorn: We'll have to build up the entire Kotlin object, yes, but we don't have to read the entire YAML source stream into memory as a string first - we can stream that and start parsing it without all of it in memory.

PR #168, opened 2021-08-11
Make faster duplicate keys checking

RSBat:
My use case involves a big file with a top level map with a bit over 40 thousand keys. Parsing it took two minutes on my laptop. After my changes parsing time went down to 10 seconds.


charleskorn:
Looks great, thanks @RSBat. Not quite sure why I was doing it that way before...

Issue #27, opened 2020-08-18
Break YamlReadingTest into smaller parts


charleskorn:
This file has become huge (2000+ lines 😱), making it difficult to navigate and slow to work with in IntelliJ.

russellbanks:
Kotest doesn't support multiplatform nested tests due to limitations in Javascript, so the tests in commonTest won't actually run on all platforms unless they are refactored to not be nested.

PR #170, opened 2021-08-14
Detect and report all duplicated keys

### LINKED ISSUE #169: Accumulate duplicated keys
In case of multiple key duplicates I would like to see information about all of them and not first one encountered. Would it be possible to make it a configurable behaviour? I can make a PR if it is something you see fitting for the library.

MiSikora:
This is binary and source incompatible change as exceptions hierarchy and structure is changed. If library users rely on  path properties of `YamlException` they will not find them in that type anymore. I decided to change it this way to make sure that everyone still catches all exceptions as `DuplicateKeysException` does not extend previous `YamlException` (now `SinglePathYamlException`).

I created a separate test suite due to #27.

Closes #169

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
@@ -267,7 +267,7 @@ private sealed class YamlMapLikeInputBase(map: YamlMap, context: SerializersModu
     protected fun <T> fromCurrentValue(action: YamlInput.() -> T): T {
         try {
             return action(currentValueDecoder)
-        } catch (e: YamlException) {
+        } catch (e: SinglePathYamlException) {
             if (currentlyReadingValue) {
                 throw InvalidPropertyValueException(propertyName, e.message, e.path, e)
             } else {
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
@@ -143,26 +143,18 @@ public data class YamlList(val items: List<YamlNode>, override val path: YamlPat
 }
 
 public data class YamlMap(val entries: Map<YamlScalar, YamlNode>, override val path: YamlPath) : YamlNode(path) {
+    private val encounteredKeys = mutableMapOf<String, YamlScalar>()
+    private val _duplicates = mutableMapOf<YamlPath, List<YamlPath>>()
+    internal val duplicates: Map<YamlPath, List<YamlPath>> = _duplicates
+
     init {
-        val keys = entries.keys.sortedWith { a, b ->
-            val lineComparison = a.location.line.compareTo(b.location.line)
-
-            if (lineComparison != 0) {
-                lineComparison
+        entries.keys.forEach { key ->
+            val encounteredKey = encounteredKeys[key.content]
+            if (encounteredKey != null) {
+                _duplicates[encounteredKey.path] = _duplicates[encounteredKey.path].orEmpty() + key.path
             } else {
-                a.location.column.compareTo(b.location.column)
+                encounteredKeys[key.content] = key
             }
-        }
-
-        val encounteredKeys = mutableMapOf<String, YamlScalar>()
-        keys.forEach { key ->
-            val duplicate = encounteredKeys[key.content]
-
-            if (duplicate != null) {
-                throw DuplicateKeyException(duplicate.path, key.path, key.contentToString())
-            }
-
-            encounteredKeys[key.content] = key
         }
     }
 
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingDuplicatesTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingDuplicatesTest.kt
@@ -1,0 +1,198 @@
+package com.charleskorn.kaml
+
+import ch.tutteli.atrium.api.fluent.en_GB.message
+import ch.tutteli.atrium.api.fluent.en_GB.notToThrow
+import ch.tutteli.atrium.api.fluent.en_GB.toBe
+import ch.tutteli.atrium.api.fluent.en_GB.toThrow
+import ch.tutteli.atrium.api.verbs.expect
+import kotlinx.serialization.decodeFromString
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+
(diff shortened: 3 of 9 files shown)


src/jvmMain/kotlin/com/charleskorn/kaml/YamlNodeReader.kt
```diff
@@ -194,6 +198,37 @@ internal actual class YamlNodeReader(
...
+                is YamlScalar, is YamlNull -> emptyList()
+            }
+        }
+
+        return if (nestedNodes.isNotEmpty()) {
+            accumulateDuplicateKeys(nestedNodes, duplicates)
+        } else {
+            duplicates
+        }
+    }
```
MiSikora, line 230: This iteration could be avoided but I didn't want `YamlMap` to accept an additional parameter for collecting duplicates.

src/jvmMain/kotlin/com/charleskorn/kaml/YamlNodeReader.kt
```diff
@@ -194,6 +198,37 @@ internal actual class YamlNodeReader(
...
 
+    private fun ensureNoDuplicateKeys(node: YamlNode) {
+        accumulateDuplicateKeys(listOf(node))
+            .takeIf { it.isNotEmpty() }
+            ?.mapValues { (_, duplicates) -> duplicates.sortedWith(compareBy({ it.endLocation.line }, { it.endLocation.column }))  }
+            ?.toSortedMap(compareBy { it.endLocation.line })
+            ?.let { throw DuplicateKeysException(it) }
+    }
+
+    private tailrec fun accumulateDuplicateKeys(
```
MiSikora, line 209: This function and iteration over all nodes could be avoided if something like `YamlNodeFactory` and make it its responsibility to calculate duplicates (and possibly do other stuff for different node types). However, I didn't want to change too much in the code structure.

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
```diff
@@ -267,7 +267,7 @@ private sealed class YamlMapLikeInputBase(map: YamlMap, context: SerializersModu
     protected fun <T> fromCurrentValue(action: YamlInput.() -> T): T {
         try {
             return action(currentValueDecoder)
-        } catch (e: YamlException) {
```
charleskorn, line 270: Should this stay as `YamlException`?

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
```diff
@@ -143,27 +143,19 @@ public data class YamlList(val items: List<YamlNode>, override val path: YamlPat
 }
 
 public data class YamlMap(val entries: Map<YamlScalar, YamlNode>, override val path: YamlPath) : YamlNode(path) {
-    init {
-        val keys = entries.keys.sortedWith { a, b ->
-            val lineComparison = a.location.line.compareTo(b.location.line)
+    private val encounteredKeys = mutableMapOf<String, YamlScalar>()
```
charleskorn, line 146: Is there a reason why we wouldn't move these into the `init` block and just immediately throw the exception in there?

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingDuplicatesTest.kt
```diff
@@ -0,0 +1,198 @@
...
+            val yaml = """
+                key1: {}
+                key1: {}
+            """.trimIndent()
+
+            it("throws an exception with a correct message") {
+                expect({ parser.decodeFromString<MapStructure>(yaml) }).toThrow<DuplicateKeysException> {
+                    message.toBe(
+                        """
+                            Found duplicates of key 'key1' defined at line 1, column 1:
```
charleskorn, line 37: Perhaps the error message would be clearer if it read as something like this:

```
Key 'key1' is only allowed once, but is defined multiple times:
- at line 1, column 1
- at line 2, column 1
```

What do you think?

charleskorn (changes requested):
Thanks for the PR. I have a couple of questions and suggestions, please see my comments.

charleskorn:
Just had an idea that might make this change less invasive - we could continue to use the "old" `YamlException`, including its path and location information as the base for `DuplicateKeysException`. The path and location information would refer to the parent map that contains the duplicate keys, and then `DuplicateKeysException` could contain information about the duplicate keys, similar to what's in this PR. 

So if we had this input YAML:

```yaml
some:
  thing:
    key1: blah
    key2: foo
    key1: bar
    key2: baz
```

We'd end up with a `DuplicateKeysException` with the path of `some.thing`, line 2, column 8, and then duplicates for `key1` on lines 3 and 5, and duplicates for `key2` on lines 4 and 6.

I don't think we need to report all duplicates in all locations throughout the tree. Doing this would be quite different behaviour to all the other exceptions kaml currently throws, as kaml currently just throws an exception on the first issue, rather than trying to aggregate all of them.

How does that sound?

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
MiSikora: Based on the current code only `SinglePathYamlException` can be thrown here. But it probably would be more future–proof to catch 'em all and do more checks inside. But if you'd go with not traversing the whole tree the root exception could stay the same so this change would be reverted.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
MiSikora: If exception is thrown immediately there is there is no way to traverse the whole tree and accumulate duplicates.

However, after reading [your comment](https://github.com/charleskorn/kaml/pull/170#issuecomment-898995605) I believe it is something you'd prefer.

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingDuplicatesTest.kt
MiSikora: Yes, this looks better. I didn't have a good idea for the message.

MiSikora:
> I don't think we need to report all duplicates in all locations throughout the tree. Doing this would be quite different behaviour to all the other exceptions kaml currently throws, as kaml currently just throws an exception on the first issue, rather than trying to aggregate all of them.
> 
> How does that sound?

In that case it is not something I was hoping for so it would be better to close this PR as I'll need to fork it or make a local copy in the project I'm currently writing.

Issue #245, opened 2022-02-26
Support for YAML block scalars


peterfigure:
### Describe the problem you'd like to solve

Working on a DSL for github actions, would be really nice if a multiline string or list of strings gets rendered as a block scalar, eg: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun

It makes for much easier debugging & readability than `\n` separated strings

### Describe the solution you'd like

https://yaml-multiline.info/ literal block scalars using the `|` symbol

charleskorn:
Resolved by #247.

PR #247, opened 2022-02-27
Add quoted scalar and multi line Style configurable support

nefilim:
https://github.com/charleskorn/kaml/issues/245

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -81,7 +81,7 @@ public enum class SingleLineStringStyle {
     SingleQuoted;
 
     public val multiLineStringStyle: MultiLineStringStyle
-        get() = when(this) {
+        get() = when (this) {
             DoubleQuoted -> MultiLineStringStyle.DoubleQuoted
             SingleQuoted -> MultiLineStringStyle.SingleQuoted
         }
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -41,8 +41,8 @@ public data class YamlConfiguration constructor(
     internal val encodingIndentationSize: Int = 2,
     internal val breakScalarsAt: Int = 80,
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
-    internal val quotedScalarStyle: QuotedScalarStyle = QuotedScalarStyle.DoubleQuoted,
-    internal val multiLineStringStyle: MultiLineStringStyle = quotedScalarStyle.multiLineStringStyle,
+    internal val singleLineScalarStyle: SingleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
+    internal val multiLineStringStyle: MultiLineStringStyle = singleLineScalarStyle.multiLineStringStyle,
 )
 
 public enum class PolymorphismStyle {
@@ -76,7 +76,7 @@ public enum class MultiLineStringStyle {
     SingleQuoted,
 }
 
-public enum class QuotedScalarStyle {
+public enum class SingleLineStringStyle {
     DoubleQuoted,
     SingleQuoted;
 
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -42,7 +42,7 @@ public data class YamlConfiguration constructor(
     internal val breakScalarsAt: Int = 80,
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
     internal val multiLineStringStyle: MultiLineStringStyle = MultiLineStringStyle.DoubleQuoted,
-    internal val stringStyle: StringStyle = StringStyle.DoubleQuoted,
+    internal val quotedScalarStyle: QuotedScalarStyle = QuotedScalarStyle.DoubleQuoted,
 )
 
 public enum class PolymorphismStyle {
@@ -76,7 +76,7 @@ public enum class MultiLineStringStyle {
     SingleQuoted,
 }
 
-public enum class StringStyle {
+public enum class QuotedScalarStyle {
     DoubleQuoted,
     SingleQuoted,
 }
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -40,7 +40,9 @@ public data class YamlConfiguration constructor(
     internal val polymorphismPropertyName: String = "type",
     internal val encodingIndentationSize: Int = 2,
     internal val breakScalarsAt: Int = 80,
-    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block
+    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
+    internal val multiLineStringStyle: MultiLineStringStyle = MultiLineStringStyle.DoubleQuoted,
+    internal val stringStyle: StringStyle = StringStyle.DoubleQuoted,
 )
 
 public enum class PolymorphismStyle {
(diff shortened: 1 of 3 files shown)


src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -77,14 +77,17 @@ internal class YamlOutput(
...
+            }
         }
     }
 
-    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) = emitQuotedScalar(enumDescriptor.getElementName(index))
+    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) = emitQuotedScalar(enumDescriptor.getElementName(index), configuration.quotedScalarStyle.scalarStyle)
 
     private fun emitPlainScalar(value: String) = emitScalar(value, ScalarStyle.PLAIN)
-    private fun emitQuotedScalar(value: String) = emitScalar(value, ScalarStyle.DOUBLE_QUOTED)
+    private fun emitQuotedScalar(value: String, scalarStyle: ScalarStyle = ScalarStyle.DOUBLE_QUOTED) = emitScalar(value, scalarStyle)
```
charleskorn, line 90: Is the default value needed?

```suggestion
    private fun emitQuotedScalar(value: String, scalarStyle: ScalarStyle) = emitScalar(value, scalarStyle)
```

charleskorn (changes requested):
Changes so far look good to me. Could you please add some tests for this? Once they're in place I'll be happy to merge this.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -40,7 +40,9 @@ public data class YamlConfiguration constructor(
     internal val polymorphismPropertyName: String = "type",
     internal val encodingIndentationSize: Int = 2,
     internal val breakScalarsAt: Int = 80,
-    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block
+    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
+    internal val multiLineStringStyle: MultiLineStringStyle = MultiLineStringStyle.DoubleQuoted,
```
charleskorn, line 44: Should this default to matching `quotedScalarStyle`? So if the user sets `quotedScalarStyle`, then `mutliLineStringStyle` uses that value unless there's an explicit value given for `multiLineStringStyle`.

```suggestion
    internal val multiLineStringStyle: MultiLineStringStyle = quotedScalarStyle, // This won't compile due to the different types, but hopefully this makes the idea clear.
```

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
nefilim: it's used below for the emitting the type tag, I can remove the default value here and move it down there, or should that honour the `quotedScalarStyle` also?

charleskorn: Ah, I'd missed that usage. I'd remove the default value and make it explicit in the caller.

nefilim:
haven't noticed any existing test for `YamlOutput` .. am I missing them? Should I add to `JvmYamlWritingTest`?

charleskorn:
> haven't noticed any existing test for `YamlOutput` .. am I missing them? Should I add to `JvmYamlWritingTest`?

`YamlWritingTest` would be the best place for these tests I think.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -40,7 +40,9 @@ public data class YamlConfiguration constructor(
     internal val polymorphismPropertyName: String = "type",
     internal val encodingIndentationSize: Int = 2,
     internal val breakScalarsAt: Int = 80,
-    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block
+    internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
+    internal val quotedScalarStyle: QuotedScalarStyle = QuotedScalarStyle.DoubleQuoted,
```
charleskorn, line 44: Should we call this `singleLineStringStyle` instead? That might make the distinction between this and `multiLineStringStyle` clearer.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -67,3 +69,20 @@ public enum class SequenceStyle {
...
+public enum class QuotedScalarStyle {
+    DoubleQuoted,
+    SingleQuoted;
+
+    public val multiLineStringStyle: MultiLineStringStyle
+        get() = when(this) {
+            DoubleQuoted -> MultiLineStringStyle.DoubleQuoted
+            SingleQuoted -> MultiLineStringStyle.SingleQuoted
+        }
+}
```
charleskorn, line 88: Should we just have one type to cover both of these? There's no harm in allowing someone to emit single line strings with the literal style.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
nefilim: this is what I had originally but it's used for characters and enums also, your call

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
nefilim: I would REALLY prefer this not to be the case, it really reduces the readability of the produced YAML and it almost doubles in length of the document when every single strings is in block literal format

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
charleskorn: Let's go with `singleLineStringStyle`.

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
```diff
@@ -28,12 +28,79 @@ import java.io.ByteArrayOutputStream
...
+                    configuration = YamlConfiguration(
+                        multiLineStringStyle = MultiLineStringStyle.DoubleQuoted,
+                    )
+                )) {
+                    val output = ByteArrayOutputStream()
+                    encodeToStream(String.serializer(), "hello\nworld\nhow are | you?\n", output)
+                    expect(output.toString(Charsets.UTF_8)).toEqual(
+                        """"hello\nworld\nhow are | you?\n"
+                            |
+                        """.trimMargin()
```
charleskorn, line 68:
```suggestion
                        """
                            |"hello\nworld\nhow are | you?\n"
                            |
                        """.trimMargin()
```

charleskorn: This might make the intention a bit clearer (both here and in the other tests)

charleskorn:
> I've added some tests, not sure why a new line is being added to every string?

Not 100% sure on this, but it's existing behaviour, so let's not add more to this PR. If it's concerning you, feel free to open another issue or raise a PR to fix it.

> ps. highly recommend you check out https://kotest.io/

I'm a huge fan of Kotest, just haven't had time to replace Spek with Kotest in here. PRs are always welcome :)

peterfigure:
there's one spotless formatting violation left in the tests, not familiar with this plugin, can you ignore it perhaps?

charleskorn:
> there's one spotless formatting violation left in the tests, not familiar with this plugin, can you ignore it perhaps?

Are you referring to this issue?

```
> The following files had format violations:
      src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
          @@ -52,7 +52,7 @@
           ·····························hello
           ·····························world
           ·····························how·are·|·you?
          -························
          +
           ························""".trimIndent()
           ····················)
           ················}
  Run './gradlew :spotlessApply' to fix these violations.
```

If so, running `./gradlew spotlessApply` should fix the issue. It's complaining about the fact that that line is just whitespace characters.

charleskorn:
Thanks @nefilim and @peterfigure!

PR #248, opened 2022-02-28
Add generic `decodeFromStream`/`encodeToStream` extension functions

### LINKED ISSUE #246: Add generic extension functions for `decodeFromStream`/`encodeToStream`

For string (de)serialization, you can use the [generic `decodeFromString` / `encodeToString` extension functions](https://github.com/Kotlin/kotlinx.serialization/blob/v1.3.2/core/commonMain/src/kotlinx/serialization/SerialFormat.kt#L84-L97) without having to explicitly specify a serializer. Since kaml's `Yaml` class [implements the `StringFormat` interface](https://github.com/charleskorn/kaml/blob/0.40.0/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt#L38-L41), this is already possible:

```kt
val myMap = Yaml.default.decodeFromString<Map<String, Int>>(myString)
val string = Yaml.default.encodeToString(myMap)
```

When decoding from a stream, this is not possible, so the serializer has to be passed explicitly, making the code harder to read:

```kt
val myMap = Yaml.default.decodeFromStream(MapSerializer(String.serializer(), Int.serializer()), myStream)
```

Introduce extension functions that read the serializer from `serializersModule.serializer()` (just like the implementation in [kotlinx.serialization.SerialFormat](https://github.com/Kotlin/kotlinx.serialization/blob/v1.3.2/core/commonMain/src/kotlinx/serialization/SerialFormat.kt#L84-L97)):

```kt
inline fun <reified T> Yaml.decodeFromStream(stream: InputStream): T =
    decodeFromStream(serializersModule.serializer(), stream)

inline fun <reified T> Yaml.encodeToStream(value: T, stream: OutputStream) =
    encodeToStream(serializersModule.serializer(), value, stream)
```

_No response_

### Additional context

_No response_

FloEdelmann:
Fixes #246.

Note that this was done via GitHub's web interface, so this exact change is untested. I have tested the `decodeFromStream` extension function locally in a downstream project, though.

The implementation (along with the KDoc comments) mirrors the one in [kotlinx.serialization.SerialFormat](https://github.com/Kotlin/kotlinx.serialization/blob/v1.3.2/core/commonMain/src/kotlinx/serialization/SerialFormat.kt#L84-L97).

### DIFF
--- a/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
+++ b/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
@@ -22,6 +22,7 @@ import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.StringFormat
+import kotlinx.serialization.serializer
 import kotlinx.serialization.modules.EmptySerializersModule
 import kotlinx.serialization.modules.SerializersModule
 import org.snakeyaml.engine.v2.api.StreamDataWriter
--- a/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
+++ b/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
@@ -22,9 +22,9 @@ import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.StringFormat
-import kotlinx.serialization.serializer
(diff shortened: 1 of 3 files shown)


charleskorn:
Yep, looks like the change is not quite right - could you please take a look?

FloEdelmann:
Should work now (but still untested).

charleskorn:
@FloEdelmann, just ran the build again and it's still failing. Could I suggest that you run the build locally yourself and fix any issues it identifies?

FloEdelmann:
Fair enough. I just ran the build and check locally and both passed.

charleskorn:
Looking good now, thanks for your contribution @FloEdelmann!

PR #286, opened 2022-05-15
Added Configuration Support for Ignoring Empty Document

DasLixou:
This adds `ignoreEmptyDocument` to the Configuration (default false)

When true, it doesnt throw an `EmptyYamlDocumentException` when the document is empty.

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -43,6 +43,7 @@ public data class YamlConfiguration constructor(
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
     internal val singleLineStringStyle: SingleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
     internal val multiLineStringStyle: MultiLineStringStyle = singleLineStringStyle.multiLineStringStyle,
+    internal val ignoreEmptyDocument: Boolean = false,
 )
 
 public enum class PolymorphismStyle {
--- a/src/jvmMain/kotlin/com/charleskorn/kaml/YamlParser.kt
+++ b/src/jvmMain/kotlin/com/charleskorn/kaml/YamlParser.kt
@@ -26,8 +26,8 @@ import org.snakeyaml.engine.v2.scanner.StreamReader
 import java.io.Reader
 import java.io.StringReader
 
-internal class YamlParser(reader: Reader) {
-    internal constructor(source: String) : this(StringReader(source))
+internal class YamlParser(reader: Reader, ignoreEmptyDocument: Boolean = false) {
+    internal constructor(source: String, ignoreEmptyDocument: Boolean = false) : this(StringReader(source), ignoreEmptyDocument)
 
     private val dummyFileName = "DUMMY_FILE_NAME"
     private val loadSettings = LoadSettings.builder().setLabel(dummyFileName).build()
@@ -37,7 +37,7 @@ internal class YamlParser(reader: Reader) {
     init {
         consumeEventOfType(Event.ID.StreamStart, YamlPath.root)
 
-        if (peekEvent(YamlPath.root).eventId == Event.ID.StreamEnd) {
+        if (peekEvent(YamlPath.root).eventId == Event.ID.StreamEnd && !ignoreEmptyDocument) {
             throw EmptyYamlDocumentException("The YAML document is empty.", YamlPath.root)
         }
 
--- a/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
+++ b/src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
@@ -49,7 +49,7 @@ public actual class Yaml(
     }
 
     private fun <T> decodeFromReader(deserializer: DeserializationStrategy<T>, source: Reader): T {
-        val parser = YamlParser(source)
+        val parser = YamlParser(source, configuration.ignoreEmptyDocument)
         val reader = YamlNodeReader(parser, configuration.extensionDefinitionPrefix)
         val rootNode = reader.read()
         parser.ensureEndOfStreamReached()


src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -43,6 +43,7 @@ public data class YamlConfiguration constructor(
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
     internal val singleLineStringStyle: SingleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
     internal val multiLineStringStyle: MultiLineStringStyle = singleLineStringStyle.multiLineStringStyle,
+    internal val ignoreEmptyDocument: Boolean = false,
```
charleskorn, line 46: This name might be a little clearer:

```suggestion
    internal val allowReadingEmptyDocument: Boolean = false,
```

src/jvmMain/kotlin/com/charleskorn/kaml/YamlParser.kt
```diff
@@ -26,8 +26,8 @@ import org.snakeyaml.engine.v2.scanner.StreamReader
 import java.io.Reader
 import java.io.StringReader
 
-internal class YamlParser(reader: Reader) {
-    internal constructor(source: String) : this(StringReader(source))
+internal class YamlParser(reader: Reader, ignoreEmptyDocument: Boolean = false) {
```
charleskorn, line 29:
```suggestion
internal class YamlParser(reader: Reader, allowEmptyDocument: Boolean = false) {
```

src/jvmMain/kotlin/com/charleskorn/kaml/YamlParser.kt
```diff
@@ -26,8 +26,8 @@ import org.snakeyaml.engine.v2.scanner.StreamReader
...
+    internal constructor(source: String, ignoreEmptyDocument: Boolean = false) : this(StringReader(source), ignoreEmptyDocument)
```
charleskorn, line 30:
```suggestion
    internal constructor(source: String, allowEmptyDocument: Boolean = false) : this(StringReader(source), ignoreEmptyDocument)
```

src/jvmMain/kotlin/com/charleskorn/kaml/YamlParser.kt
```diff
@@ -37,7 +37,7 @@ internal class YamlParser(reader: Reader) {
     init {
         consumeEventOfType(Event.ID.StreamStart, YamlPath.root)
 
-        if (peekEvent(YamlPath.root).eventId == Event.ID.StreamEnd) {
+        if (peekEvent(YamlPath.root).eventId == Event.ID.StreamEnd && !ignoreEmptyDocument) {
```
charleskorn, line 40:
```suggestion
        if (peekEvent(YamlPath.root).eventId == Event.ID.StreamEnd && !allowEmptyDocument) {
```

charleskorn (changes requested):
Thanks for the PR @DasLixou! 

Two things:
* I have a minor name change suggestion (see comments).
* Could you please add a test for this?

DasLixou:
Ok I’ll add a test later :)

charleskorn:
> @charleskorn What have i done wrong?
> 
> grafik

I believe that's just a bug in IntelliJ - does it compile OK from the command line? Try `./gradlew check` (this will also run the tests).

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
```diff
@@ -42,5 +44,18 @@ class JvmYamlReadingTest : DescribeSpec({
...
+        describe("reading an empty yaml without throwing an error") {
+            val input = ""
+            val bytes = ByteArrayInputStream(input.toByteArray(Charsets.UTF_8))
+
+            it("expect throwing an error because it's an empty document") {
+                shouldThrowExactly<EmptyYamlDocumentException> { Yaml.default.decodeFromStream<String>(bytes) }
+            }
+
+            it("expect ignoring empty document because of configuration") {
+                shouldNotThrowAny { Yaml(configuration = YamlConfiguration(allowReadingEmptyDocument = true)).decodeFromStream<String>(bytes) }
```
charleskorn, line 57: In this scenario (reading a non-nullable string), what would you expect to be returned from `decodeFromStream`? Can we make an assertion that shows the expected behaviour?

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
```diff
@@ -42,5 +44,18 @@ class JvmYamlReadingTest : DescribeSpec({
                 result shouldBe 123
             }
         }
+
+        describe("reading an empty yaml without throwing an error") {
```
charleskorn, line 48: These changes aren't specific to the JVM, but this file is for tests that only apply to the JVM. Could you please move them to the `commonTest` source set?

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
```diff
@@ -42,5 +44,18 @@ class JvmYamlReadingTest : DescribeSpec({
...
+            val input = ""
+            val bytes = ByteArrayInputStream(input.toByteArray(Charsets.UTF_8))
+
+            it("expect throwing an error because it's an empty document") {
+                shouldThrowExactly<EmptyYamlDocumentException> { Yaml.default.decodeFromStream<String>(bytes) }
```
charleskorn, line 53: It would be good to add tests to cover a couple of other scenarios as well. For example, reading lists, maps and Kotlin classes.

charleskorn (changes requested):
Just a few suggestions on the tests to verify this behaves as expected and then this should be good to go!

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
DasLixou: I tried it but the problem is that the common Kotlin version of YAML only has the default variable. If there were an expected function decodeFromStream or so it would work..

DasLixou:
I dont know how, but my last braincell made it working :D - After looking at it a second time it just totally make sense.
Now ill add the other test cases

DasLixou:
PROBLEM: When the document is empty, i cant get an object out of YamlInput. Is there any way to get the default object of an Deserialization? @charleskorn

src/jvmMain/kotlin/com/charleskorn/kaml/YamlNodeReader.kt
```diff
@@ -19,13 +19,8 @@
 package com.charleskorn.kaml
```
charleskorn, line 19: Could you please revert the formatting changes in this file? It makes it very difficult to see what's really changed and what is just a formatting change.

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
charleskorn: The common Kotlin version has a `decodeFromString` method that could be used in the tests.

charleskorn:
> PROBLEM: When the document is empty, i cant get an object out of YamlInput. Is there any way to get the default object of an Deserialization? @charleskorn

This probably requires changes in kotlinx.serialization. What does the built-in JSON format do if you try to deserialize an empty JSON stream to an object like this?

DasLixou:
Screw it. If anyone has more braincells, just go ahead and fork DasLixou/kaml if you want. I'll just check if the file is empty before running Kaml. This system is little more complicated.

charleskorn:
Given the issues discussed above, I'm going to close this as "won't fix".

PR #287, opened 2022-05-31
Added experimental comment support on serialization

slava110:
Hi. I've created an annotation to add comments for properties on serialization. This PR could be improved so I'm waiting for any suggestions (or PR rejection)
Some questions:
- Should `@YamlComment` annotation be marked as `@ExperimentalSerializationApi`?
- Do you want configuration value to disable comment serialization?
- Are tests good enough or there should be more?

P.s.: Edits by maintainers should be allowed too

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -43,7 +43,6 @@ public data class YamlConfiguration constructor(
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
     internal val singleLineStringStyle: SingleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
     internal val multiLineStringStyle: MultiLineStringStyle = singleLineStringStyle.multiLineStringStyle,
-    internal val serializeComments: Boolean = false,
 )
 
 public enum class PolymorphismStyle {
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -43,6 +43,7 @@ public data class YamlConfiguration constructor(
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
     internal val singleLineStringStyle: SingleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
     internal val multiLineStringStyle: MultiLineStringStyle = singleLineStringStyle.multiLineStringStyle,
+    internal val serializeComments: Boolean = false,
 )
 
 public enum class PolymorphismStyle {
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
@@ -898,7 +898,7 @@ class YamlWritingTest : DescribeSpec({
                         # Testing
                         # multiline
                         test: "justTest"
-                        """.trimIndent()
+                    """.trimIndent()
                 }
             }
         }
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
@@ -32,7 +32,6 @@ import com.charleskorn.kaml.testobjects.polymorphicModule
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.DescribeSpec
 import io.kotest.matchers.shouldBe
-import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.PolymorphicSerializer
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.builtins.ListSerializer
@@ -888,16 +887,16 @@ class YamlWritingTest : DescribeSpec({
         }
 
         describe("handling comments") {
-            context("comments in Java object") {
+            context("comments in kotlin object") {
                 val input = SimpleStructureWithComments("objName", 73, "justTest")
 
                 it("is written") {
                     Yaml.default.encodeToString(SimpleStructureWithComments.serializer(), input) shouldBe """
                         name: "objName"
-                        #Cool int
+                        # Cool int
                         myInt: 73
-                        #Testing
-                        #multiline
+                        # Testing
+                        # multiline
                         test: "justTest"
                         """.trimIndent()
                 }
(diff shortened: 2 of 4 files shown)


slava110:
Also wanted to add annotation parameter for users in case they don't want to trim indent in multiline comments but default annotation parameters are [probably broken](https://github.com/Kotlin/kotlinx.serialization/issues/1956)

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -43,6 +43,7 @@ public data class YamlConfiguration constructor(
     internal val sequenceStyle: SequenceStyle = SequenceStyle.Block,
     internal val singleLineStringStyle: SingleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
     internal val multiLineStringStyle: MultiLineStringStyle = singleLineStringStyle.multiLineStringStyle,
+    internal val serializeComments: Boolean = false,
```
charleskorn, line 46: Is there a use case you're thinking of where someone would want to disable this? (I'm wondering if we can just always serialize comments and reduce the complexity of having this be configurable.)

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -90,6 +92,23 @@ internal class YamlOutput(
...
+            if(comment != null) {
+                // Is multiline comment?
+                if(comment.indexOfAny(charArrayOf('\n', '\r')) > -1)
+                    comment = comment.trimIndent()
+
+                for (line in comment.lineSequence()) {
+                    emitter.emit(CommentEvent(CommentType.BLOCK, line, Optional.empty(), Optional.empty()))
+                }
+            }
+        }
```
charleskorn, line 110: Perhaps this could be moved to a separate method to help convey the intention more clearly and help keep `encodeElement` short and clear?

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -90,6 +92,23 @@ internal class YamlOutput(
...
+        if(configuration.serializeComments) {
+            var comment = descriptor.getElementAnnotations(index)
+                .filterIsInstance<YamlComment>()
+                .firstOrNull()
+                ?.comment
+
+            if(comment != null) {
+                // Is multiline comment?
+                if(comment.indexOfAny(charArrayOf('\n', '\r')) > -1)
+                    comment = comment.trimIndent()
```
charleskorn, line 104: In the interests of consistency, should we just always `trimIndent` the comment? Then any leading spaces are always removed, regardless of whether the comment is one or multiple lines.

src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
```diff
@@ -885,6 +886,23 @@ class YamlWritingTest : DescribeSpec({
...
         }
+
+        describe("handling comments") {
+            context("comments in Java object") {
+                val input = SimpleStructureWithComments("objName", 73, "justTest")
+
+                it("is written") {
+                    Yaml.default.encodeToString(SimpleStructureWithComments.serializer(), input) shouldBe """
+                        name: "objName"
+                        #Cool int
```
charleskorn, line 897: Should there be a space between the `#` and the comment content perhaps? Or should this be configurable? What do you think?

charleskorn (commented):
Thanks for the PR @slava110!

> - Should @YamlComment annotation be marked as @ExperimentalSerializationApi?

I think it's OK to not mark it as experimental.

> - Do you want configuration value to disable comment serialization?

See below.

> - Are tests good enough or there should be more?

If we allow enabling or disabling comments, then we should at least test the behaviour when comments are enabled and disabled.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
slava110: Yeah, I think you're right, configuration option is redundant. Thanks for pointing it out! And sorry, it's my first actual PR, so I'm making some mistakes on the way

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
slava110: Yes, I can do it if you want

src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
slava110: Not sure. Like an additional `@YamlComment` annotation parameter `prefix`? I don't think it's needed but I can add it if you want

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
charleskorn: No worries at all, congratulations on your first PR 🎉

src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
charleskorn: I think it's fairly safe to default to adding a space - if anyone asks for something different in the future, we can make it configurable.

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -142,6 +146,23 @@ internal class YamlOutput(
...
+        if(commentAnno != null) {
+            for (line in commentAnno.lines) {
+                emitter.emit(CommentEvent(CommentType.BLOCK, " $line", Optional.empty(), Optional.empty()))
+            }
+            // FIXME: add `inline` annotation parameter and uncomment when kotlin serialization issue 1956 will be resovled
+            // (default parameters on annotations marked with @SerialInfo will throw an exception currently)
+            // Issue link: https://github.com/Kotlin/kotlinx.serialization/issues/1956
+            /*if(commentAnno.inline.isNotEmpty())
+                emitter.emit(CommentEvent(CommentType.IN_LINE, " ${ commentAnno.inline }", Optional.empty(), Optional.empty()))*/
+        }
```
charleskorn, line 163:
```suggestion
        if (commentAnno == null) {
            return
        }
        
        for (line in commentAnno.lines) {
            emitter.emit(CommentEvent(CommentType.BLOCK, " $line", Optional.empty(), Optional.empty()))
        }
```

charleskorn (changes requested):
This is looking good. I've got one minor nitpick (see comment below) and ktlint has requested some changes (run `./gradlew spotlessApply` to fix these automatically).

Once those are resolved, I'm happy to merge this - thanks again @slava110!

slava110:
Not `spotless plugin` but yeah
An interesting experience for me. Thanks for guiding me in this journey c:

PR #289, opened 2022-06-05
feat: support for text block

Colerar:
#287 works well with oneline comment and multiple argument.

But do not support for text block.

Text block will cause runtime-error. 

Like this:

```kotlin
// Source code
@YamlComment(
"""
Max pool size
Default and Recommended Value:
  SQLite: 1, because SQLite has no multi-connection support
  PostgresQL: Processor Threads, like 12 for 6-core machine
    see more: https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing
""")
```

```yaml
# Yaml file
#
            Max pool size
            Default and Recommended Value:
              SQLite: 1, because SQLite has no multi-connection support
              PostgresQL: Processor Threads, like 12 for 6-core machine
                see more: https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing
```

I add a new Annotation `YamlCommentTrim` to fix this issue. I don't know if it's best practice. At first, I try to add a `trimType` parameter with default value for `YamlComment`, but caused build-error.

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlComment.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlComment.kt
@@ -18,6 +18,8 @@
 
 package com.charleskorn.kaml
 
+import com.charleskorn.kaml.TrimType.TRIM_INDENT
+import com.charleskorn.kaml.TrimType.TRIM_MARGIN
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerialInfo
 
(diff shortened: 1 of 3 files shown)


charleskorn:
I'd prefer to fix this with a property on the `YamlComment` annotation, but this is blocked by https://github.com/Kotlin/kotlinx.serialization/issues/1956.

Let's see if the kotlinx.serialization team can fix that bug quickly - if not, I'm happy to consider something like this.

slava110:
I mean, instead of using multiline string you can use multiple string arguments...
If it's not convenient I think it'd be better to remove `vararg` at all then. I just thought it doesn't matter for users and it's easier to process in code

Colerar:
I think it's ok if I use kaml from beginning. But my project is planning to migrate from [yamlkt](https://github.com/him188/yamlkt), and it supports text block.

I also think it's a more natural way to edit multi-line comments by text block.

Compare:

```
@Comment(
  "this is line1",
  "", // maybe a empty line for better format
  "this is line2",
)
```

Colerar:
I think it's ok if I use kaml from beginning. But my project is planning to migrate from [yamlkt](https://github.com/him188/yamlkt), and it supports text block.

I also think it's a more natural way to edit multi-line comments by text block.

Compare:

```
@Comment(
  "this is line1",
  "", // maybe a empty line for better format
  "this is line2",
  "this is line3",
)
```

```
@Comment(
  """
  this is line1

  this is line2
  this is line3
  """
)
```

slava110:
*Huh*
https://github.com/Him188/yamlkt/blob/master/yamlkt/src/commonMain/kotlin/net.mamoe.yamlkt/YamlComment.kt#L57
Okay then

Colerar:
> [https://github.com/Him188/yamlkt/blob/master/yamlkt/src/commonMain/kotlin/net.mamoe.yamlkt/YamlComment.kt#L57](https://github.com/Him188/yamlkt/blob/master/yamlkt/src/commonMain/kotlin/net.mamoe.yamlkt/YamlComment.kt?rgh-link-date=2022-06-09T09%3A40%3A10Z#L57)

That's why I want to migrate to kaml. YamlKt is unstable and its only benefit is multiplatform. But I do not need it with JVM-only project.

> If it's not convenient I think it'd be better to remove `vararg` at all then. I just thought it doesn't matter for users and it's easier to process in code

Yes, we can remove `vararg` at all if compatibility is not considered (kaml is in `0.x` version)

But if compatibility is a important factor, we can provide 2 ways at the same time.

----

If this pr is still blocked by kotlinx.serialization#1956 after few days. I think we can invoke `trimIndent()` as default. But add `trimType` option after the issue fixed. Because text block even do not work in current implmentation.

Issue #231, opened 2022-01-17
Ability to serialize/deserialize unstructured data


sjshuck:
### Describe the problem you'd like to solve

Sometimes I need to be able to break the glass and serialize/deserialize YAML without having well-structured POJOs/KOJOs.

### Describe the solution you'd like

I'd like a canonical "vanilla" representation of YAML.  It does not have to implement all YAML features (e.g., SnakeYAML chokes on tags in its vanilla codec implementation, and I'd be fine with that behavior).  It could be a codec for `YamlNode` or `Any` or a new class created for this purpose.

This would bring it in line with other languages' YAML implementations such as Python's and Ruby's.

### Describe alternatives you've considered

A combination of SnakeYAML and Moshi.

### Additional context

My context is also multi-document YAML streams from Kubernetes, just like #54 's case.

charleskorn:
Thanks for the report @sjshuck.

To clarify your request, are you looking for something like [kotlinx.serialization's `JsonElement`](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#parsing-to-json-element)? Or are you looking for a generic YAML parser independent of kotlinx.serialization?

If the latter, that is currently out of scope for kaml - it specifically targets kotlinx.serialization, and is not a general-purpose YAML parser for Kotlin. If you need a general-purpose YAML parser, I'd suggest using a library like snakeyaml-engine (which is what kaml uses under the hood).

sjshuck:
Hey, thank _you_ for the response.

I think I'm looking for the former.  In addition to `JsonElement`, there's the helper method `Json.parseToJsonElement`.  I was looking for something like `Yaml.default.parseToYamlNode`.  In fact, I'd be fine with `Yaml.default.parseToJsonElement` or with a deserialization strategy for `JsonElement`, if there are concerns about making one for `YamlNode` (YAML itself being too complex for a faithful two-way generic codec, perhaps).

Regarding it being in scope, I get that snakeyaml and `kotlinx.serialization` are distinct levels of abstraction and in a sense distinct concerns.  But may I suggest that exposing the underlying repr via a codec for the repr's type is a pretty common pattern among serialization/deserialization libraries in various languages, and I always appreciate seeing it there for development, testing, and exploration.  Here, it would save me a dependency on snakeyaml.  A nice-to-have.

charleskorn:
[...]

PR #296, opened 2022-06-16
Add parsing to YamlNode

kitterion:
Add the ability to deserialize yaml data without a rigid schema.

I've recently found myself needing to deserialize some yaml data where I wished I could use something like [parseToJsonElement](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json/parse-to-json-element.html). Unfortunately, there was no way to get a YamlNode to do manual parsing, so here I add parseToYamlNode.
I also add a few convenience functions similar to what kotlinx.serialization has for json.

Partly addresses #231, specifically the deserialization part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,30 @@ val result = Yaml.default.encodeToString(Team.serializer(), input)
 println(result)
 ```
 
+### Parsing into YamlNode
+
+It is possible to parse a string or an InputStream directly into a YamlNode, for example
+the following code prints `Cindy`.
+```kotlin
+val input = """
+        leader: Amy
+        members:
+          - Bob
+          - Cindy
+          - Dan
+    """.trimIndent()
+
+val result = Yaml.default.parseToYamlNode(input)
+
+println(
+    result
+        .yamlMap.get<YamlNode>("members")!!
+        .yamlList[1]
+        .yamlScalar
+        .content
+)
+```
+
 ## Referencing kaml
 
 Add the following to your Gradle build script:
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
@@ -116,6 +116,8 @@ public data class YamlList(val items: List<YamlNode>, override val path: YamlPat
         return this.items.zip(other.items).all { (mine, theirs) -> mine.equivalentContentTo(theirs) }
     }
 
+    public operator fun get(index: Int): YamlNode = items[index]
+
     override fun contentToString(): String = "[" + items.joinToString(", ") { it.contentToString() } + "]"
 
     override fun withPath(newPath: YamlPath): YamlList {
@@ -247,4 +249,23 @@ public data class YamlTaggedNode(val tag: String, val innerNode: YamlNode) : Yam
     override fun withPath(newPath: YamlPath): YamlNode = this.copy(innerNode = innerNode.withPath(newPath))
 
     override fun toString(): String = "tagged '$tag': $innerNode"
+}
+
+public val YamlNode.yamlScalar: YamlScalar
+    get() = this as? YamlScalar ?: error(this, "YamlScalar")
+
+public val YamlNode.yamlNull: YamlNull
+    get() = this as? YamlNull ?: error(this, "YamlNull")
+
+public val YamlNode.yamlList: YamlList
+    get() = this as? YamlList ?: error(this, "YamlList")
+
+public val YamlNode.yamlMap: YamlMap
+    get() = this as? YamlMap ?: error(this, "YamlMap")
+
+public val YamlNode.yamlTaggedNode: YamlTaggedNode
+    get() = this as? YamlTaggedNode ?: error(this, "YamlTaggedNode")
+
+private fun error(node: YamlNode, expectedType: String): Nothing {
+    throw IncorrectTypeException("Element is not $expectedType", node.path)
 }
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlListTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlListTest.kt
@@ -18,6 +18,7 @@
 
 package com.charleskorn.kaml
 
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.DescribeSpec
 import io.kotest.matchers.shouldBe
 
@@ -72,6 +73,29 @@ class YamlListTest : DescribeSpec({
                 it("indicates that they are not equivalent") {
                     list.equivalentContentTo(YamlMap(emptyMap(), list.path)) shouldBe false
                 }
+            }
+        }
+
+        describe("getting elements from a list") {
+            val firstItemPath = YamlPath.root.withListEntry(0, Location(4, 5))
+            val secondItemPath = YamlPath.root.withListEntry(1, Location(6, 7))
+
+            val list = YamlList(
+                listOf(
+                    YamlScalar("item 1", firstItemPath),
+                    YamlScalar("item 2", secondItemPath)
+                ),
+                YamlPath.root
+            )
+
+            it("returns element") {
+                list[0] shouldBe YamlScalar("item 1", firstItemPath)
+                list[1] shouldBe YamlScalar("item 2", secondItemPath)
+            }
+
+            it("throws IndexOutOfBoundsException") {
+                shouldThrow<IndexOutOfBoundsException> { list[2] }
+                shouldThrow<IndexOutOfBoundsException> { list[10] }
             }
         }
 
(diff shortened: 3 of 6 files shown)


src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
```diff
@@ -248,3 +250,22 @@ public data class YamlTaggedNode(val tag: String, val innerNode: YamlNode) : Yam
...
+    get() = this as? YamlNull ?: error(this, "YamlNull")
+
+public val YamlNode.yamlList: YamlList
+    get() = this as? YamlList ?: error(this, "YamlList")
+
+public val YamlNode.yamlMap: YamlMap
+    get() = this as? YamlMap ?: error(this, "YamlMap")
+
+public val YamlNode.yamlTaggedNode: YamlTaggedNode
+    get() = this as? YamlTaggedNode ?: error(this, "YamlTaggedNode")
```
charleskorn, line 267: Perhaps these would be clearer named like `asScalar`, `asNull`, `asList` etc.?

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
```diff
@@ -248,3 +250,22 @@ public data class YamlTaggedNode(val tag: String, val innerNode: YamlNode) : Yam
...
+
+private fun error(node: YamlNode, expectedType: String): Nothing {
+    throw IncorrectTypeException("Element is not $expectedType", node.path)
```
charleskorn, line 270: It would be good to include the actual type of the node in addition to what was expected (eg. "Expected element to be a list, but is a scalar")

src/commonTest/kotlin/com/charleskorn/kaml/YamlListTest.kt
```diff
@@ -75,6 +76,29 @@ class YamlListTest : DescribeSpec({
...
+
+            val list = YamlList(
+                listOf(
+                    YamlScalar("item 1", firstItemPath),
+                    YamlScalar("item 2", secondItemPath)
+                ),
+                YamlPath.root
+            )
+
+            it("returns element") {
```
charleskorn, line 91: For these two tests, it would be good to describe the situation in addition to the desired behaviour (eg. "if the index is within the bounds of the list's size" or "if the index is outside the bounds of the list's size").

src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
```diff
@@ -49,15 +49,24 @@ public actual class Yaml(
...
 
         val input = YamlInput.createFor(rootNode, serializersModule, configuration, deserializer.descriptor)
         return input.decodeSerializableValue(deserializer)
     }
 
+    public fun parseToYamlNode(string: String): YamlNode = parseToYamlNodeFromReader(StringReader(string))
+
+    public fun parseToYamlNode(source: InputStream): YamlNode = parseToYamlNodeFromReader(InputStreamReader(source))
+
+    private fun parseToYamlNodeFromReader(source: Reader): YamlNode {
```
charleskorn, line 62: It would be nice to continue the pattern of all parsing methods using the verb "decode" in their names - so these methods could be something like `decodeToYamlNode`. Or did you deliberately choose to use "parse"?

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
```diff
@@ -42,5 +42,23 @@ class JvmYamlReadingTest : DescribeSpec({
                 result shouldBe 123
             }
         }
+
+        describe("parsing into a YamlNode from a string") {
+            val input = "123"
+            val result = Yaml.default.parseToYamlNode(input)
+
+            it("successfully deserializes values from a string") {
+                result.yamlScalar.toInt() shouldBe 123
```
charleskorn, line 51: These two tests are testing a lot more than just parsing to a `YamlNode` - they're also testing the `yamlScalar` extension property and integer conversion. Is there a way to narrow the scope of the test to just focus on the parsing? 

(You've already added tests to cover the other aspects, which is great by the way 🎉)

src/commonTest/kotlin/com/charleskorn/kaml/YamlNodeTest.kt
```diff
@@ -0,0 +1,60 @@
...
+            Triple("YamlList", YamlNode::yamlList, YamlList(emptyList(), path)),
+            Triple("YamlMap", YamlNode::yamlMap, YamlMap(emptyMap(), path)),
+            Triple(
+                "YamlTaggedNode", YamlNode::yamlTaggedNode,
+                YamlTaggedNode("tag", YamlScalar("tagged_scalar", path))
+            ),
+        )
+
+        cases.forEach { (fromType, _, value) ->
+            cases.forEach { (toType, method, _) ->
```
charleskorn, line 43: This test has a little too much magic in it for my liking - I'd prefer the cases were more explicitly laid out.

README.md
```diff
@@ -53,6 +53,30 @@ val result = Yaml.default.encodeToString(Team.serializer(), input)
...
+          - Bob
+          - Cindy
+          - Dan
+    """.trimIndent()
+
+val result = Yaml.default.parseToYamlNode(input)
+
+println(
+    result
+        .yamlMap.get<YamlNode>("members")!!
```
charleskorn, line 73: If the `get` call is changed to `get<YamlList>()`, does that remove the need for the `yamlList` call?

charleskorn (changes requested):
Thanks for the PR @kitterion!

The general approach looks good to me, I have a few suggestions about some of the details.

src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
kitterion: This was indeed a deliberate decision to mirror [parseToJsonElement](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json/parse-to-json-element.html) from kotlinx.serialization. I did think about naming them `decode...` but it seemed to me like being closer to kotlinx.serialization.json is more important.

Looking at it more closely now, it stems from the fact that encoding/decoding happens with `@Serializable` classes, and here no `@Serializable` classes are involved. Their [documentation](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#parsing-to-json-element) puts it well:

> A string can be parsed into an instance of [JsonElement](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-element/index.html) with the [Json.parseToJsonElement](https://kotlin.github.io/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json/parse-to-json-element.html) function. It is called neither decoding nor deserialization because none of that happens in the process. It just parses a JSON and forms an object representing it

Here my suggestion is to take the same path and reserve both `decodeFromYamlNode` and `encodeToYamlNode` for `YamlNode` <-> `@Serializable` conversion.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
kitterion: This was also deliberate to mirror the extension properties from kotlinx.serialization.json https://github.com/Kotlin/kotlinx.serialization/blob/43d5f7841fc744b072a636b712e194081456b5ba/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt#L148-L174

src/jvmMain/kotlin/com/charleskorn/kaml/Yaml.kt
charleskorn: Makes sense to me - thanks for clarifying.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
charleskorn: Makes sense to me - thanks for clarifying.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
kitterion: Done.

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlReadingTest.kt
kitterion: Done.

src/commonTest/kotlin/com/charleskorn/kaml/YamlNodeTest.kt
kitterion: I'm not sure if that's what you meant, but i unrolled these nested loops and separated the successful tests from the tests that throw.

README.md
kitterion: Changed it to `get<YamlList>`.

charleskorn (approved):
Looks good to me, thanks @kitterion (and sorry for the slow review!)

PR #302, opened 2022-07-03
Initial implementation of Yaml*Serializers

pschichtel:
this is still very much broken and there a bunch of open questions, but this could be a viable starting point. This was basically adapted from the Json equivalents. The Json module seems to have more support for this in the Encoder/Decoder directly.

Someone with more knowledge of the inner workings of this module should have a look at this and also at the test failures I'm seeing. 

The idea is to be able to decode and re-encode with these serializers and get the exact same value back, that's what the tests try to verify.

Since I'm switching over to a different approach for my project, I will not be needing this myself, so depending on how much work is left here I may not be able to continue working on this actively.


charleskorn:
Thanks for submitting this @pschichtel. Are you able to expand on the issues you ran into and what you think could be done to fix them?

pschichtel:
I'm not even sure if accessing the Input instances in the way I did is correct. In one case I seemed to receive a YamlMapInput where I would have expected to receive a YamlScalarInput.

I think if you run the unit test I added you'll see most of the issues I had

PR #337, opened 2022-10-04
Update SnakeYAML repository URL

FloEdelmann:
* https://bitbucket.org/asomov/snakeyaml-engine consists of only an outdated wiki (see the "Changes" page there, it lists v2.3 from last year as latest version).
* https://bitbucket.org/snakeyaml/snakeyaml-engine has the source code, issues and an up-to-date wiki. I linked to the wiki home page directly, as it's not very discoverable from the repository home page.


charleskorn:
Good catch, thanks @FloEdelmann!

PR #369, opened 2023-01-13
Add decoding from YamlNode

dellisd:
This adds a method on `YamlInput` that allows for decoding a serializable value from a `YamlNode`. 

The use case for this kind of functionality is the ability to manually interpret the structure of a YAML document and then delegate the deserialization of `YamlNode`s to other serializer implementations.

For example, in the test case I added we have this document:
```yaml 
keyA:
    host: A
keyB:
    host: B
```
which is a map of `Database` objects. Perhaps we want to decode this as a `List<Database>` instead of a `Map<String, Database>` (because reasons) and this enables us to write a custom serializer that looks at each `YamlNode` entry on the `YamlMap` and then delegates the deserialization of each `Database` object to the `Database` serializer rather than having to manually decode the rest of the Yaml structure.

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
@@ -116,6 +116,12 @@ public sealed class YamlInput(
 
     public abstract fun getCurrentLocation(): Location
     public abstract fun getCurrentPath(): YamlPath
+
+    public fun <T : Any?> decodeFromYamlNode(
+        node: YamlNode,
+        deserializer: DeserializationStrategy<T>,
+    ): T =
+        createFor(node, serializersModule, configuration, deserializer.descriptor).decodeSerializableValue(deserializer)
 }
 
 @OptIn(ExperimentalSerializationApi::class)
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
@@ -32,49 +32,50 @@ import kotlinx.serialization.modules.SerializersModule
 @OptIn(ExperimentalSerializationApi::class)
 public sealed class YamlInput(
     public val node: YamlNode,
+    public val yaml: Yaml,
     override var serializersModule: SerializersModule,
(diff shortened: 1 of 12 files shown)


charleskorn:
Thanks for the contribution @dellisd. 

I'm wondering if this would be better suited as a method on `Yaml` instead of on `YamlInput`, as the method doesn't strictly require a `YamlInput` instance. I'm imagining a method like `decodeFromNode(node, serializer)`. What do you think?

Also, could you please add a test case for the scenario where the serializer is decoding something that isn't the root of the document? There might be an edge case there that I'd like to check.

dellisd:
I agree adding a method to `Yaml` makes sense, but I think the `YamlInput` class will also need to be modified to expose the current `Yaml` instance so that custom settings are preserved and new `Yaml` instances don't need to be allocated during deserialization. This is similar to what is done with the [`JsonDecoder` implementation](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/JsonDecoder.kt#L55-L58).

I can make these changes, and I'll add that test case.

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
```diff
@@ -32,49 +32,50 @@ import kotlinx.serialization.modules.SerializersModule
 @OptIn(ExperimentalSerializationApi::class)
 public sealed class YamlInput(
     public val node: YamlNode,
+    public val yaml: Yaml,
```
charleskorn, line 35: Why is this `yaml: Yaml` property required? Is this so that custom deserializers can access it?

charleskorn (commented):
Overall looks good, just have a couple of questions...

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
dellisd: Yes, that's correct. I explained above that the Json implementations in kotlinx.serialization expose the `Json` instances in the same way.

charleskorn (approved):

charleskorn:
Thanks @dellisd!

PR #400, opened 2023-03-15
Add YamlNamingStrategy

### LINKED ISSUE #398: Add Yaml Naming Strategies

There is no naming strategy implementation, so `@SerialName` has to be applied to every variable in a data class:

```kotlin
@SerialName("PackageIdentifier") val packageIdentifier: String,
@SerialName("PackageVersion") val packageVersion: String
```

A YamlNamingStrategy in the YamlConfiguration.

```kotlin
configuration = YamlConfiguration(
    yamlNamingStrategy = YamlNamingStrategy.PascalCase // CamelCase, SnakeCase, etc
)
```

`@SerialName` on every variable

### Additional context

KotlinX Serialization recently added Json naming strategies in [1.5.0](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.5.0-RC):

> A long-awaited feature (https://github.com/Kotlin/kotlinx.serialization/issues/33) is available in this release.
A new interface, JsonNamingStrategy and Json configuration property namingStrategy allow defining a transformation that is applied to all properties' names serialized by a Json instance.
There's also a predefined implementation for the most common use case: Json { namingStrategy = JsonNamingStrategy.SnakeCase }.
Check out the https://github.com/Kotlin/kotlinx.serialization/pull/2111 for more details and documentation.

```kotlin
Json {
    namingStrategy = JsonNamingStrategy.SnakeCase // Predefined
}
```

```kotlin
Json {
    namingStrategy = JsonNamingStrategy { descriptor: SerialDescriptor, elementIndex: Int, serialName: String ->
        // Able to define custom naming strategy
    }
}
```
https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-naming-strategy

russellbanks:
Adds YamlNamingStrategy from #398 with pre-defined snake_case, kebab-case, PascalCase, and camelCase.

The SnakeCase implementation has been taken and modified from [JsonNamingStrategy.kt](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt) in KotlinX Serialization.

## Usage

```kotlin
YamlConfiguration(
    yamlNamingStrategy = YamlNamingStrategy.PascalCase  // CamelCase, SnakeCase, KebabCase
),
````
#### Example data class:
```kotlin
@Serializable
data class Data(
   val propertyName: String,
   val otherName: String
)
```
#### Example Yaml:
```yaml
PropertyName: "value"
OtherName: "value"
```
This will successfully serialize and deserialize without the need for `@SerialName` on every value in the data class.

#### Converts to lowercase and capitalises first letter:
```kotlin
YamlConfiguration(
    yamlNamingStrategy = { it.lowercase().replaceFirstChar(Char::titlecase) }
),
````
#### Converts to uppercase
```kotlin
YamlConfiguration(
    yamlNamingStrategy = String::uppercase
),
````

### Note

As `@SerialName` is resolved at compile time rather than runtime, using a YamlNamingStrategy will ignore the casing of all `@SerialName` annotations.

Closes #398

### CODE REVIEW

**Comment by russellbanks** [comment]
@charleskorn can you help or provide any pointers for deserializing the input and applying the naming strategy? I think it may need a [custom serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/JsonElementSerializers.kt#L29), or where SerialName is overriden with the Yaml Naming Strategy applied.

File: src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt [hunk 1]
```diff
@@ -96,8 +96,8 @@ internal class YamlOutput(
 
     override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) = emitQuotedScalar(enumDescriptor.getElementName(index), configuration.singleLineStringStyle.scalarStyle)
 
-    private fun emitPlainScalar(value: String) = emitScalar(value, ScalarStyle.PLAIN)
-    private fun emitQuotedScalar(value: String, scalarStyle: ScalarStyle) = emitScalar(value, scalarStyle)
+    private fun emitPlainScalar(value: String) = emitScalar(value, ScalarStyle.PLAIN, configuration.yamlNamingStrategy)
+    private fun emitQuotedScalar(value: String, scalarStyle: ScalarStyle) = emitScalar(value, scalarStyle, configuration.yamlNamingStrategy)
```
charleskorn, line 100: Should this only apply to field names? At present, we'll apply the naming strategy to all field names and also all scalar values.

charleskorn:
Thank you for the PR @russellbanks!

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
russellbanks: I made the mistake of applying this to the values, not the serial names. I need to look into this to see how to change the serial names. The algorithms to transform a string into a case type work well though so it's just finding a way to apply it.

kenyee: IMHO this should only be applied to field names

kenyee: Just to confirm, this will ignore the @SerializeName?

I'd like to be able to use the same data model classes for JSON and YAML serialization (I've done this with Jackson before).
Alternative is @SerializeName should be able to specify json/yaml variants

kenyee:
any idea when this will be merged?
Could use this on a project I'm working on :-)

charleskorn:
> any idea when this will be merged? Could use this on a project I'm working on :-)

This is currently an incomplete draft - it'll be merged when it's ready :)

If you'd like to see it merged sooner, I'm sure @russellbanks would love some help.

russellbanks:
It's the linking up of the algorithms to the serial names that would be helpful to have some pointers for :)

charleskorn:
> It's the linking up of the algorithms to the serial names that would be helpful to have some pointers for :)

[This](https://github.com/charleskorn/kaml/blob/647350e10297eca1137e9e4360d930b6530f1335/src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt#L106) is where property names are emitted - so that should be the place to perform the transformation.

russellbanks:
@charleskorn I've done it correctly this time, with it being applied to property names and only the property names.

src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
russellbanks: I think it makes the most sense for this to be applied to all items, but `@SerialName` be an override to this

russellbanks: `@SerialName` is resolved at compile time,there isn't a way to know if it was applied to a specific value or not, so strategies will ignore it. This is how it works with KotlinX Serialization's Json one too. Naming strategies should cover most use cases apart from when there is inconsistent naming.

russellbanks:
@charleskorn This is ready for review I think!

One thing that I should note is that applying a naming strategy will ignore all @SerialName annotations as that is resolved at compile time rather than runtime, meaning there's no way to know if it was applied or not.

src/commonMain/kotlin/com/charleskorn/kaml/YamlObjectInput.kt
```diff
@@ -36,11 +36,27 @@ internal class YamlObjectInput(map: YamlMap, yaml: Yaml, context: SerializersMod
...
+
+            val pairedPropertyNames = (0 until descriptor.elementsCount)
+                .asSequence()
+                .map(descriptor::getElementName)
+                .map { elementName ->
+                    val yamlName = configuration.yamlNamingStrategy?.serialNameForYaml(elementName) ?: elementName
+                    elementName to yamlName
+                }
+
+            val fieldDescriptorIndex = pairedPropertyNames
```
charleskorn, line 48: This might be simpler if `pairedPropertyNames` was a map of converted names to their field index

src/commonMain/kotlin/com/charleskorn/kaml/YamlObjectInput.kt
```diff
@@ -36,11 +36,27 @@ internal class YamlObjectInput(map: YamlMap, yaml: Yaml, context: SerializersMod
...
+                .find { (_, convertedName) -> propertyName == convertedName }
+                ?.let { (descriptorName, _) -> descriptor.getElementIndex(descriptorName) }
+                ?: CompositeDecoder.UNKNOWN_NAME
 
             if (fieldDescriptorIndex == CompositeDecoder.UNKNOWN_NAME) {
                 if (configuration.strictMode) {
-                    throwUnknownProperty(propertyName, currentKey.path, descriptor)
+                    throw UnknownPropertyException(
+                        propertyName,
+                        pairedPropertyNames.map { (_, convertedName) -> convertedName }.toSet(),
```
charleskorn, line 57: If `pairedPropertyNames` was a map of converted names to their field indices, then this could be `pairedPropertyNames.keys`

src/commonMain/kotlin/com/charleskorn/kaml/YamlObjectInput.kt
```diff
@@ -36,11 +36,27 @@ internal class YamlObjectInput(map: YamlMap, yaml: Yaml, context: SerializersMod
 
             val currentEntry = entriesList[nextIndex]
             currentKey = currentEntry.key
-            val fieldDescriptorIndex = descriptor.getElementIndex(propertyName)
+
+            val pairedPropertyNames = (0 until descriptor.elementsCount)
```
charleskorn, line 40: It would be nice if we didn't have to build this `pairedPropertyNames` for every call to `decodeElementIndex` - is there any way to avoid this?

src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
```diff
@@ -297,6 +297,81 @@ class YamlWritingTest : DescribeSpec({
...
+
+            context("deserializing a camelCase serial name using YamlNamingStrategy.CamelCase") {
+                val output = Yaml(
+                    configuration = YamlConfiguration(yamlNamingStrategy = YamlNamingStrategy.CamelCase),
+                ).decodeFromString(NamingStrategyTestData.serializer(), "serialName: value")
+
+                it("correctly serializes into the data class") {
+                    output shouldBe NamingStrategyTestData("value")
+                }
+            }
```
charleskorn, line 372: Deserialisation tests should go in `YamlReadingTest.kt`.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNamingStrategy.kt
```diff
@@ -0,0 +1,97 @@
...
+    public fun serialNameForYaml(serialName: String): String
+
+    public companion object Builtins {
+        /**
+         * A [YamlNamingStrategy] that converts property names to snake_case (lowercase words separated by underscores).
+         */
+        public val SnakeCase: YamlNamingStrategy = object : YamlNamingStrategy {
+            override fun serialNameForYaml(serialName: String): String = serialName.toDelimitedCase('_')
+
+            override fun toString(): String = "com.charleskorn.kaml.YamlNamingStrategy.SnakeCase"
```
charleskorn, line 31: Are the `toString()` overrides required?

src/commonMain/kotlin/com/charleskorn/kaml/YamlNamingStrategy.kt
```diff
@@ -0,0 +1,97 @@
...
+   limitations under the License.
+
+*/
+
+package com.charleskorn.kaml
+
+public fun interface YamlNamingStrategy {
+    public fun serialNameForYaml(serialName: String): String
+
+    public companion object Builtins {
```
charleskorn, line 24: Would be good to add some tests specifically for these implementations (eg. test that they behave correctly with empty strings, single characters, single words, multiple words etc.)

charleskorn (commented):
Thanks for putting this together @russellbanks! Overall it's looking good.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNamingStrategy.kt
russellbanks: Not necessarily - KotlinX Serialization did that for their Json equivalent of this class, so I did too

src/commonMain/kotlin/com/charleskorn/kaml/YamlObjectInput.kt
russellbanks: I've changed it to use lateinit and initialise the first time `decodeElementIndex` is called in https://github.com/charleskorn/kaml/pull/400/commits/21d4295d43c93b32c6619611c83c6102ac28b054

src/commonMain/kotlin/com/charleskorn/kaml/YamlNamingStrategy.kt
russellbanks: A blank serial name isn't possible, unless `@SerialName` is used. In this case, the result is:
```kotlin
@Serializable
data class EmptySerialName(@SerialName("") val name: String)
```
Expected (or perhaps an exception):
```yaml
": "value""
```
Actual:
```yaml
"?
: "value""
```

This happens even in the absence of any naming strategy as blank serial names are not expected or intended. If it's not intentional, it's out of the scope of this PR to validate that behaviour.

russellbanks:
@charleskorn I've made the requested changes! Let me know if there's anything else.

charleskorn (commented):
This is looking great, thanks again for this @russellbanks. 

I've just noticed one more case that is missing a test: could you please add a test for the case where an unknown property is included in the YAML document being read? I want to make sure the exception includes the transformed names (from the naming strategy) rather than the property in the descriptor.

russellbanks:
> This is looking great, thanks again for this @russellbanks.
> 
> I've just noticed one more case that is missing a test: could you please add a test for the case where an unknown property is included in the YAML document being read? I want to make sure the exception includes the transformed names (from the naming strategy) rather than the property in the descriptor.

Well thought! I've been able to map that but this test case fails and I'm not sure how to map the YamlMap's entries.
```kotlin
context("given some input representing an object with an unknown key, using a naming strategy") {
    val input = "oneTwoThree: something".trimIndent()

    context("parsing that input") {
        it("throws an exception with the naming strategy applied") {
            val exception = shouldThrow<UnknownPropertyException> {
                Yaml(
                    configuration = YamlConfiguration(yamlNamingStrategy = YamlNamingStrategy.SnakeCase)
                ).decodeFromString(ComplexStructure.serializer(), input)
            }

            exception.asClue {
                it.message shouldBe "Unknown property 'one_two_three'. Known properties are: boolean, byte, char, double, enum, float, int, long, nullable, short, string"
                it.line shouldBe 1
                it.column shouldBe 1
                it.propertyName shouldBe "one_two_three"
                it.validPropertyNames shouldBe setOf("boolean", "byte", "char", "double", "enum", "float", "int", "long", "nullable", "short", "string")
                it.path shouldBe YamlPath.root.withMapElementKey("one_two_three", Location(1, 1))
            }
        }
    }
}
```

This is the test case that fails
```kotlin
it.path shouldBe YamlPath.root.withMapElementKey("one_two_three", Location(1, 1))
```
```
Expected :YamlPath(segments=[com.charleskorn.kaml.YamlPathSegment$Root@1ebcc98f, MapElementKey(key=one_two_three, location=Location(line=1, column=1))])
Actual   :YamlPath(segments=[com.charleskorn.kaml.YamlPathSegment$Root@1ebcc98f, MapElementKey(key=oneTwoThree, location=Location(line=1, column=1))])
```

charleskorn:
I think the test is not quite right - the exception should contain the properties as they appear / would appear in YAML.

So this:

```kotlin
it.path shouldBe YamlPath.root.withMapElementKey("one_two_three", Location(1, 1))
```

should be:

```kotlin
it.path shouldBe YamlPath.root.withMapElementKey("oneTwoThree", Location(1, 1))
```

And the exception message should be something like `Unknown property 'oneTwoThree'. Known properties are: boolean, byte, ...`

russellbanks:
@charleskorn I've done that and I have also tested this PR by publishing to Maven Local and it works well in my project :)

charleskorn:
Thanks again for the PR @russellbanks!

PR #423, opened 2023-05-09
Replace kotlinOptions with compilerOptions

russellbanks:
In Kotlin 1.8.20, `kotlinOptions` was deprecated in favour of `compilerOptions`. This means there is type safety with specifying the JVM target now.

https://kotlinlang.org/docs/whatsnew1820.html
https://youtu.be/R1JpkpPzyBU?t=150


charleskorn (approved):
Thanks @russellbanks!

PR #444, opened 2023-07-11
Flatten Yaml List Test

russellbanks:


charleskorn:
Thanks @russellbanks!

PR #452, opened 2023-07-24
Flatten Yaml Scalar Test

russellbanks:

listOf has been replaced with [withData](https://kotest.io/docs/framework/datatesting/data-driven-testing.html#getting-started)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -74,6 +74,7 @@ kotlin {
                 implementation("io.kotest:kotest-assertions-core:5.6.2")
                 implementation("io.kotest:kotest-framework-api:5.6.2")
                 implementation("io.kotest:kotest-framework-engine:5.6.2")
+                implementation("io.kotest:kotest-framework-datatest:5.6.2")
             }
         }
 
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
@@ -20,397 +20,273 @@ package com.charleskorn.kaml
 
 import io.kotest.assertions.asClue
 import io.kotest.assertions.throwables.shouldThrow
-import io.kotest.core.spec.style.DescribeSpec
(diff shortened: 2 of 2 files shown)


russellbanks:
In Kotlin/JS equality for NaN should be done by `isNaN()`. In Kotest, this corresponds to `shouldBeNan()` so I've put tests relating to NaN separately so they work in JavaScript.

russellbanks:
@charleskorn One issue is that the tests `0x2` and `0o2` are successfully converted to a float in JavaScript (On the JVM it throws an exception).

```kotlin
val scalar = YamlScalar("0x2", index1Line2Column4Path)

val floatException = shouldThrow<YamlScalarFormatException>(scalar::toFloat) // Gets converted to 2
```

On the JVM, the test passes. On JavaScript, there is `AssertionError: Expected exception YamlScalarFormatException but no exception was thrown.`

The below functions might need extra logic to covert this edge case in JavaScript:
https://github.com/charleskorn/kaml/blob/7461bb14682cabd87a462ff54ad737f1950f9620/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt#L56 https://github.com/charleskorn/kaml/blob/7461bb14682cabd87a462ff54ad737f1950f9620/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt#L70

charleskorn:
> One issue is that the tests `0x2` and `0o2` are successfully converted to a float in JavaScript (On the JVM it throws an exception).
> 
> ...
> 
> The below functions might need extra logic to cover this edge case in JavaScript:

This makes sense to me - could you please add this extra logic with `expect` / `actual` declarations for each target?

Alternatively, should we raise this as a Kotlin bug? I'd expect those methods to behave identically across platforms.

russellbanks:
> Alternatively, should we raise this as a Kotlin bug? I'd expect those methods to behave identically across platforms.

I'm not sure with this honestly. Looking at the code in Kotlin, they just delegate the behaviour to their related platforms.

PR #458, opened 2023-08-25
Add Kotlin Native targets

russellbanks:
This adds Kotlin Native targets to Kaml:
- mingwX64
- linuxX64
- macosX64
- macosArm64

These are the ones that are [supported](https://github.com/krzema12/snakeyaml-engine-kmp/blob/main/buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform-native.gradle.kts) by SnakeYaml KMP at the moment. More can be added once they support more.

This PR does the same as #437.

Alternatively, the code can be moved into commonMain so that we would only have a common module but there are still platform specific functions for the JVM like decodeFromStream, etc, and I'm not sure on how to integrate that while the other code goes in commonMain. However, there is the consideration of the maturity of SnakeYaml KMP with putting the code into commonMain - https://github.com/charleskorn/kaml/pull/437#issuecomment-1617499408:

> Long-term, I'm totally for it as well, but for now our KMP port is yet immature and using it already for the JVM could introduce some regressions

Because this is the same as the JS code, another option is to create a non-JVM source set rather than nativeMain and jsMain.

I'm not sure why but Batect doesn't seem to like the Native tests.

It might be worth also running the CI on Windows, Linux and macOS rather than just Linux to ensure that the Windows and macOS binaries are building correctly.

@krzema12 @aSemy


charleskorn:
> I'm not sure why but Batect doesn't seem to like the Native tests.

Not sure either - I'll take a look. I'm considering removing Batect from this project altogether, so this might be the thing that pushes me to do it.

> It might be worth also running the CI on Windows, Linux and macOS rather than just Linux to ensure that the Windows and macOS binaries are building correctly.

That would be great - could you please add that? (Feel free to wait until I've removed Batect if that's easier.)

charleskorn:
>Because this is the same as the JS code, another option is to create a non-JVM source set rather than nativeMain and jsMain.

That seems like a reasonable thing to do for now - once the port is in a more stable place, we can figure out how to share the code across JVM and non-JVM platforms.

build.gradle.kts
```diff
@@ -78,24 +83,50 @@ kotlin {
             }
         }
 
-        named("jvmMain") {
+        val jvmMain by getting {
+            dependsOn(commonMain)
```
aSemy, line 87: Can you double check this is necessary? Adding `dependsOn(commonMain)` should be automatic https://kotlinlang.org/docs/multiplatform-share-on-platforms.html#share-code-on-all-platforms

> You don't need to specify any `dependsOn` relations manually:
> 
> * For all platform-specific source sets that depend on the common source set, such as `jvmMain`, `macosX64Main`, and others

```suggestion
```

aSemy:
> That seems like a reasonable thing to do for now - once the port is in a more stable place, we can figure out how to share the code across JVM and non-JVM platforms.

hey @charleskorn, I'm curious about the stability side, because it's definitely important so I'd like to have more of an idea. Have there been any issues with Kaml/JS? If you look in the Maven Central stats, has it been used a lot? If it's been used a fair amount and there are no issues, then I'd say it's stable enough! 

However, what is more likely is that if Kaml/JVM switched from SnakeYAML-Java to SnakeYaml-KMP then there might be some subtle changes in behaviour - although based on the [YAML Test Suite tests](https://github.com/krzema12/snakeyaml-engine-kmp/pull/59) I doubt it.

charleskorn:
>Have there been any issues with Kaml/JS?

Not that I'm aware of - no one has reported any issues. 

>If you look in the Maven Central stats, has it been used a lot?

I don't have access to any statistics on this unfortunately.

charleskorn:
I've removed Batect from the CI pipeline - please rebase and see if that helps with the native tests.

build.gradle.kts
```diff
@@ -78,24 +83,49 @@ kotlin {
...
+            dependsOn(commonTest)
+
             dependencies {
                 implementation("io.kotest:kotest-runner-junit5:5.6.2")
             }
         }
 
-        named("jsMain") {
+        val nonJVMMain by creating {
+            dependsOn(commonMain)
```
charleskorn, line 102: Should this be added to the `jvmMain` set as well?

charleskorn (commented):
Could you please add a mention of the experimental native targets to the readme?

Also, could you please expand the GitHub Actions workflow to run the tests on Windows and macOS hosts as well?

build.gradle.kts
russellbanks: No, to my understanding, all default source sets (the ones retrieved by `by getting`) implicitly depend on commonMain. We have to do it for nonJVMMain because we are creating it as a new source set. We then have to specify that all the non JVM source sets depend on this one. See @aSemy's [review](https://github.com/charleskorn/kaml/pull/458#discussion_r1308459399) that highlighted this to me.

There's a good example of this [here](https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#deprecated-api-for-adding-kotlin-source-sets-directly-to-the-kotlin-compilation) under *What's the best practice now?*

russellbanks:
> Also, could you please expand the GitHub Actions workflow to run the tests on Windows and macOS hosts as well?

I've done this in a separate pull request - #470

russellbanks:
The tests seem to fail when writing reports because of https://github.com/gradle/gradle/issues/21547. For now, we could just run jvmTest while the flattening of tests is worked on, as this would also reduce the length of the file names.

build.gradle.kts
charleskorn: Makes sense, thanks for clarifying.

PR #468, opened 2023-09-27
Replace manual actions/cache with gradle action

russellbanks:
The [Gradle GitHub Action](https://github.com/gradle/gradle-build-action) automatically handles cache to do with Gradle and Gradle build files for you:

> ## [Caching build state between Jobs](https://github.com/gradle/gradle-build-action#caching-build-state-between-jobs)
> The `gradle-build-action` will use the GitHub Actions cache to save and restore reusable state that may be speed up a subsequent build invocation. This includes most content that is downloaded from the internet as part of a build, as well as expensive to create content like compiled build scripts, transformed Jar files, etc.

This also fixes the renovate issue that used a commit hash for actions/checkout v4 rather than a version number.

Also, may I ask why the build file uses specific version numbers? E.g. @4.1.0 rather than @v4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - name: Check out code
-        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        uses: actions/checkout@v4.1.0
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


charleskorn:
>Also, may I ask why the build file uses specific version numbers? E.g. @4.1.0 rather than @v4

This helps keep builds more predictable - with less specific version numbers like `@v4`, there's a chance re-running the same build will use different versions of actions.

charleskorn (approved):

charleskorn:
Thanks @russellbanks!

PR #469, opened 2023-09-27
Use Java 17 in build.yml

russellbanks:
Java 18 is end-of-life. This pull request changes the build file to use Java 17 as that is the latest LTS release that has been built for Temurin. I would suggest upping this to Java 21 once Adoptium has created binaries for it and Kotlin supports Java 21.


charleskorn (approved):
Thanks @russellbanks!

PR #470, opened 2023-09-27
Run workflow on Ubuntu, Windows and macOS

russellbanks:
This runs the build workflow on ubuntu-latest, windows-latest, and macos-latest. `fail-fast` is set to false so that if the build fails on one operating system the others will not stop.

The release steps will only run on Ubuntu as to not create and release the binaries multiple times. This will need to changed (along with some changes to the build file) in #458 after this PR so that native binaries can be built and released without causing duplicate binaries. Kotlin has [documentation](https://kotlinlang.org/docs/multiplatform-publish-lib.html#avoid-duplicate-publications) on this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,11 @@ on: [push, pull_request]
 jobs:
   build:
     name: "Build, Test and Publish"
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


russellbanks:
This workflow fails on Windows because `rm -rf` (which is done to delete caches) isn't a command on Windows.

Merging #468 will resolve this.

charleskorn (commented):
Could you please split the assemble and publish steps out to their own job that waits on the build jobs to complete? Otherwise, as this stands, there's a risk that we'll publish a release while the Windows or macOS builds are failing.

PR #483, opened 2023-11-12
Fix typo in sealed class example

jleskovar-tyro:
Fix up typo - `Backend` and `Frontend` should both be of type `Server`


charleskorn:
Thanks!

charleskorn (approved):

Issue #378, opened 2023-02-05
PolymorphismStyle.None


mgroth0:
### Describe the problem you'd like to solve

I would like to serialize classes belonging to a polymorphic hierarchy as if they were not. I understand this means they can no longer be correctly deserialized, so I understand why there might be hesitation to support such a feature. Maybe it can be part of a separate API.

### Describe the solution you'd like

allowing `PolymorphismStyle.None`, which would simply do nothing when a class is polymorphic, would work for me. Alternatively, having a separate method to serialize objects without considering if they are polymorphic would be ok. 

### Describe alternatives you've considered

I am using `PolymorphismStyle.Tag` and trying to remove these tags by searching the string.

### Additional context

I may be serializing an object which will be deserialized in another program altogether, and that other program does not need the polymorphic tags.

charleskorn:
Thanks for the suggestion @mgroth0. If you're interested in submitting a PR with this change, I'd be happy to accept it.

ZacSweers:
I'll take a look at a PR

ZacSweers:
#494

PR #494, opened 2023-12-14
Introduce PolymorphismStyle.None

ZacSweers:
This implements #378 by introducing a new `None` style that omits the tag/property entirely from the output yaml.

Note this is asymmetric and reads are treated the same as `Tag` as reads require a tag.

I added a few tests to JvmYamlWritingTest for both this case and some added coverage for the existing two cases as well.

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -35,7 +35,7 @@ package com.charleskorn.kaml
  * * [sequenceBlockIndent]: number of spaces to use as indentation for sequences, if [sequenceStyle] set to [SequenceStyle.Block]
  * * [allowAnchorsAndAliases]: set to true to allow anchors and aliases when decoding YAML (defaults to `false`)
  */
-public data class YamlConfiguration constructor(
+public data class YamlConfiguration(
     internal val encodeDefaults: Boolean = true,
     internal val strictMode: Boolean = true,
     internal val extensionDefinitionPrefix: String? = null,
@@ -55,6 +55,7 @@ public data class YamlConfiguration constructor(
 public enum class PolymorphismStyle {
     Tag,
     Property,
+    None,
 }
 
 public enum class SequenceStyle {
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
@@ -70,7 +70,10 @@ public sealed class YamlInput(
             }
 
             is YamlTaggedNode -> when {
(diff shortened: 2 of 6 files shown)


src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -35,7 +35,7 @@ package com.charleskorn.kaml
  * * [sequenceBlockIndent]: number of spaces to use as indentation for sequences, if [sequenceStyle] set to [SequenceStyle.Block]
  * * [allowAnchorsAndAliases]: set to true to allow anchors and aliases when decoding YAML (defaults to `false`)
  */
-public data class YamlConfiguration constructor(
+public data class YamlConfiguration(
```
ZacSweers, line 38: This was opportunistic as it was redundant

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
```diff
@@ -63,14 +63,16 @@ public sealed class YamlInput(
...
                     PolymorphismStyle.Property -> createPolymorphicMapDeserializer(node, yaml, context, configuration)
                 }
                 else -> throw IncorrectTypeException("Expected ${descriptor.kind.friendlyDescription}, but got a map", node.path)
             }
 
             is YamlTaggedNode -> when {
-                descriptor.kind is PolymorphicKind && configuration.polymorphismStyle == PolymorphismStyle.Tag -> YamlPolymorphicInput(node.tag, node.path, node.innerNode, yaml, context, configuration)
+                descriptor.kind is PolymorphicKind && configuration.polymorphismStyle != PolymorphismStyle.Property -> {
+                    YamlPolymorphicInput(node.tag, node.path, node.innerNode, yaml, context, configuration)
+                }
```
charleskorn, line 75: I think it might be simpler to just fail if we encounter polymorphism or a tag and `PolymorphismStyle.None` is in use.

charleskorn (commented):
Thanks for the PR @ZacSweers!

Overall, this looks good to me. 

In addition to the comments below, could you please also add a test for the behaviour when reading polymorphic and non-polymorphic content while `PolymorphismStyle.None` is in use?

ZacSweers:
Will look when I'm back from vacation 👍

ZacSweers:
Done!

charleskorn (approved):
Thanks @ZacSweers!

charleskorn:
> I think there's a bug that I missed in `YamlOutput` here:
> 
> https://github.com/charleskorn/kaml/blob/2f0de209a6143d271c385475ac62b321e25f7a1e/src/jvmMain/kotlin/com/charleskorn/kaml/YamlOutput.kt#L175-L177
> 
> I think in the None case, it should just omit the tag right? If so, I can PR a quick fix

That makes sense to me - a PR with a test case and fix would be great.

PR #579, opened 2024-06-21
Add native targets

### LINKED ISSUE #232: Non-JVM targets are not supported

OptimumCode:
There is a problem with saving test results for native target - this issue will probably be resolved after #574 is merged.
There are also two tests failed (for invalid floating point with input `1e+` and `1e-` - the problem was reported to Kotlin team in [KT-69327](https://youtrack.jetbrains.com/issue/KT-69327)).

Resolves #232 

Please, let me know if I have missed something


OptimumCode:
Hi, @charleskorn. Could you please take a look at the PR?

buildSrc/src/main/kotlin/com/charleskorn/kaml/build/ConfigureAssemble.kt
```diff
@@ -33,7 +33,8 @@ fun Project.configureAssemble() {
         project.extensions.getByType<PublishingExtension>().publications.names
             .filter { it != "kotlinMultiplatform" }
             .forEach { publicationName ->
-                from(tasks.named("${publicationName}Jar"))
+                // only Jvm and JS targets have jar task. All others have *binaries task
+                tasks.findByName("${publicationName}Jar")?.let { from(it) }
```
charleskorn, line 37: This will eagerly create the `...Jar` task. Is there another way to achieve the same thing without eagerly creating these tasks? https://docs.gradle.org/current/userguide/task_configuration_avoidance.html explains this further.

build.gradle.kts
```diff
@@ -111,6 +129,12 @@ tasks.withType<KotlinCompile>().configureEach {
     }
 }
 
+tasks.withType<KotlinNativeTest> {
+    // A few test with invalid floating point number fail due to issue in Kotlin native parser
+    // https://youtrack.jetbrains.com/issue/KT-69327
+    enabled = false
+}
+
```
charleskorn, line 137: Could we conditionally disable just the affected tests rather than disabling all native tests?

build.gradle.kts
```diff
@@ -79,6 +80,23 @@ kotlin {
         }
     }
 
+    //region Native Targets
```
charleskorn, line 83: The `//region` and `//endregion` comments are not necessary.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
```diff
@@ -63,6 +63,9 @@ public data class YamlScalar(val content: String, override val path: YamlPath) :
                     content.toFloat()
                 } catch (e: NumberFormatException) {
                     throw YamlScalarFormatException("Value '$content' is not a valid floating point value.", path, content)
+                } catch (e: IndexOutOfBoundsException) {
+                    // https://youtrack.jetbrains.com/issue/KT-69327
```
charleskorn, line 67: Perhaps this would be clearer?
```suggestion
                    // Workaround for https://youtrack.jetbrains.com/issue/KT-69327
```

Could you please make the same change below as well?

charleskorn (commented):
Thanks for working on this @OptimumCode!

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: "Build, Test and Publish"
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
charleskorn, line 8: We'll need to change the workflow definition to run this job on Linux, macOS and Windows - otherwise we'll only run the macOS, iOS and iPadOS native tests.

We'll also have to figure out how to build and publish the binaries across three OSes as well.

OptimumCode: As far as I know MacOS is capable of building and running tests for all targets (for linux and windows as well). But I will recheck this just in case.
In my own project I use the same approach for publishing artifacts for all platforms and it works well. I think it should be fine here as well

buildSrc/src/main/kotlin/com/charleskorn/kaml/build/ConfigureAssemble.kt
OptimumCode: I didn't find an equivalent in Lazy task api. But I think we can do it another way:
```kotlin
if (publicationName in PUBLICATIONS_WITH_JAR) {
  from(tasks.named(...))
}
```

build.gradle.kts
OptimumCode: Sure, will do. I was waiting for #574 to be merged as it has some changes needed to run tests for native target

.github/workflows/build.yml
OptimumCode: You were right, tests for Linux and Windows targets are skipped (as well as tests for macOSX64). I will update the workflow to execute tests for each target on the corresponding runner. But the assembling and publication can be done on macOS for all targets

OptimumCode: You can find run results [here](https://github.com/OptimumCode/kaml/actions/runs/9632865172). Please, let me know if you have any suggestions on how to improve the workflow

.github/workflows/build.yml
```diff
@@ -3,9 +3,34 @@ name: Pipeline
...
+  test-macosX64:
+    uses: ./.github/workflows/test-target.yml
+    with:
+      runs-on: macos-13 # latest version with X64 processor
+      tasks: macosX64Test iosX64Test
+  test-macosArm64:
+    uses: ./.github/workflows/test-target.yml
+    with:
+      runs-on: macos-latest
+      tasks: macosArm64Test iosSimulatorArm64Test
```
charleskorn, line 25: Couldn't this be a single job with a [matrix](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude)? We could just use `jobs.*.strategy.matrix.include` to explicitly define the jobs and the tasks we want them to run.

OptimumCode: Yes, it could be. But it won't be possible to use a reusable workflow in matrix job (makes it a bit less readable in my opinion).
I will convert those jobs to a matrix job and you can compare both options

OptimumCode:
@charleskorn Should I add a concurrency group to prevent multiple workflows from running for the same PR?

.github/workflows/build.yml
```diff
@@ -2,14 +2,61 @@ name: Pipeline
...
+  GRADLE_OPTS: "-Dorg.gradle.internal.launcher.welcomeMessageEnabled=false"
+
 jobs:
-  build:
-    name: "Build, Test and Publish"
-    runs-on: ubuntu-latest
+  test-matrix:
+    strategy:
+      matrix:
+        os: [ 'ubuntu-latest', 'windows-latest', 'macos-13', 'macos-latest' ]
```
charleskorn, line 12: This line is not necessary - the `include` list below is sufficient.

.github/workflows/build.yml
```diff
@@ -2,14 +2,61 @@ name: Pipeline
...
 on: [push, pull_request]
 
+env:
+  GRADLE_OPTS: "-Dorg.gradle.internal.launcher.welcomeMessageEnabled=false"
+
 jobs:
-  build:
-    name: "Build, Test and Publish"
-    runs-on: ubuntu-latest
+  test-matrix:
```
charleskorn, line 9: We should add a name for this job - something like `Test (${{matrix.ox}})` perhaps?

charleskorn:
> @charleskorn Should I add a concurrency group to prevent multiple workflows from running for the same PR?

Yep, sounds like a good idea.

.github/workflows/build.yml
OptimumCode: Sound good, will do

.github/workflows/build.yml
```diff
@@ -2,14 +2,65 @@ name: Pipeline
 
 on: [push, pull_request]
 
+env:
+  GRADLE_OPTS: "-Dorg.gradle.internal.launcher.welcomeMessageEnabled=false"
+
+concurrency:
+  cancel-in-progress: true
+  group: build-${{ github.event.pull_request.number || github.event.after }}
```
OptimumCode, line 10: This allows only one workflow run per PR and, in the meantime, does not affect the push event (because `github.event.after` will be unique for every commit). @charleskorn does it look okay?

charleskorn: LGTM, thanks.

charleskorn (approved):
Thanks @OptimumCode!

OptimumCode:
[KT-69327](https://youtrack.jetbrains.com/issue/KT-69327) should be resolved in Kotlin 2.1.0 based on the latest comment in the issue

PR #582, opened 2024-06-24
Correct handling of `SerialKind.CONTEXTUAL` in `YamlInput` class

### LINKED ISSUE #568: IllegalStateException: Must call beginStructure() and use returned Decoder when using `@Contextual`
### Describe the bug

`yaml.decodeFromString()` throws an `IllegalStateException: Must call beginStructure() and use returned Decoder` when using the `@Contextual` annotation on a property 

### Reproduction repo

```kotlin
import com.charleskorn.kaml.Yaml
import kotlinx.serialization.Contextual
import kotlinx.serialization.KSerializer
import kotlinx.serialization.Serializable
import kotlinx.serialization.descriptors.PrimitiveKind
import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
import kotlinx.serialization.descriptors.SerialDescriptor
import kotlinx.serialization.encodeToString
import kotlinx.serialization.encoding.Decoder
import kotlinx.serialization.encoding.Encoder
import kotlinx.serialization.modules.SerializersModule
import java.time.LocalDate
import kotlin.test.Test
import kotlin.test.assertEquals

class ConfigTest {
    private val yaml: Yaml = Yaml(
        serializersModule = SerializersModule {
            contextual(LocalDate::class, LocalDateSerializer)
        },
    )

    @Test
    fun `Test serialization and deserialization`() {
        val config = Config(
            dateOfBirth = LocalDate.parse("1980-01-01"),
        )

        val yamlString = yaml.encodeToString(config)
        assertEquals(config, yaml.decodeFromString(Config.serializer(), yamlString))
    }
}

@Serializable
data class Config(
    @Contextual
    val dateOfBirth: LocalDate,
)


object LocalDateSerializer : KSerializer<LocalDate> {
    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("LocalDate", PrimitiveKind.STRING)

    override fun serialize(encoder: Encoder, value: LocalDate) {
        encoder.encodeString(value.toString())
    }

    override fun deserialize(decoder: Decoder): LocalDate {
        val string = decoder.decodeString()
        return LocalDate.parse(string)
    }
}
```

### Steps to reproduce

Run the test `Test serialization and deserialization`

### Expected behaviour

The data class
```kotlin
@Serializable
data class Config(
    @Contextual
    val dateOfBirth: LocalDate,
)
```
is correctly encoded and decoded.

### Actual behaviour

The data class `Config` is correctly encoded (`dateOfBirth: "1980-01-01"`) but, when decoded it throws an `IllegalStateException`.

### Version information

```plain text
`"com.charleskorn.kaml:kaml:0.60.0"`
```

### Any other information

Using
```kotlin
@Serializable
data class Config(
    @Serializable(with = LocalDateSerializer::class)
    val dateOfBirth: LocalDate,
)
```
instead of
```kotlin
@Serializable
data class Config(
    @Contextual
    val dateOfBirth: LocalDate,
)
```
works around the issue but it would be nice to be able to use the `@Contextual` annotation.

OptimumCode:
Hi, @charleskorn.

This PR corrects how `YamlInput` handles the serialization of fields marked with `@Contextual` annotation.
Instead of forcing the serializer to call `beginStructure` it extracts the actual definition and creates the appropriate `YamlInput` based on that definition where applicable.

Please, take a look at the PR once you have time. Let me know if you have any remarks/suggestions regarding the change

Resolves #568

### DIFF
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
@@ -2079,7 +2079,7 @@ class YamlReadingTest : FlatFunSpec({
                     override val descriptor: SerialDescriptor
                         get() = String.serializer().descriptor
 
-                    override fun deserialize(decoder: Decoder): Inner = Inner(decoder.decodeString())
+                    override fun deserialize(decoder: Decoder): Inner = Inner("from context serializer: ${decoder.decodeString()}")
                     override fun serialize(encoder: Encoder, value: Inner) = throw UnsupportedOperationException()
                 }
 
@@ -2093,7 +2093,7 @@ class YamlReadingTest : FlatFunSpec({
                 val result = parser.decodeFromString(Container.serializer(), input)
 
                 test("deserializes it using the dynamically installed serializer") {
-                    result shouldBe Container(Inner("this is the input"))
+                    result shouldBe Container(Inner("from context serializer: this is the input"))
                 }
             }
 
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
@@ -2079,7 +2079,7 @@ class YamlReadingTest : FlatFunSpec({
                     override val descriptor: SerialDescriptor
                         get() = String.serializer().descriptor
 
-                    override fun deserialize(decoder: Decoder): Inner = Inner("from context serializer")
+                    override fun deserialize(decoder: Decoder): Inner = Inner(decoder.decodeString())
                     override fun serialize(encoder: Encoder, value: Inner) = throw UnsupportedOperationException()
                 }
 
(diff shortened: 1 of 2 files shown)


src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -2079,7 +2079,7 @@ class YamlReadingTest : FlatFunSpec({
                     override val descriptor: SerialDescriptor
                         get() = String.serializer().descriptor
 
-                    override fun deserialize(decoder: Decoder): Inner = Inner("from context serializer")
+                    override fun deserialize(decoder: Decoder): Inner = Inner(decoder.decodeString())
```
charleskorn, line 2082: How about something like this, so that we can be confident the contextual serializer was used?

```suggestion
                    override fun deserialize(decoder: Decoder): Inner = Inner("from context serializer: " + decoder.decodeString())
```

charleskorn (commented):
Thanks for fixing this @OptimumCode! 

I have one comment about the test, but otherwise this looks good to me.

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
OptimumCode: There is no other serializer that can be used because `Inner` class is not annotated with `@Serializable`. But I don't see any harm if we modify the test as you suggested)

charleskorn (approved):
Thanks @OptimumCode!

PR #584, opened 2024-06-27
Update tests to run for JS targets

OptimumCode:
Hi, this PR allows running tests for JS targets.

The old version of tests had the following problems:
- It is incorrect to compare NaN to another NaN using `==` operator. For Java `Float.NaN == Float.NaN` returns `false` (but `Objects.equals(Float.NaN, Float.NaN)` returns `true `, which is why tests work for Java). For JS it is [a bit different](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN).
  > NaN compares unequal (via [==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality), [!=](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Inequality), [===](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality), and [!==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality)) to any other value — including to another NaN value.

  For NaN comparison, `equals` call was replaced with `isNaN()` method call.

- The hex and octal numbers are valid floating point numbers for JS (because under the hood Kotlin delegates conversion to JS dynamic cast). This is the `toDouble` implementation for JS platform
  ```kotlin
  public actual fun String.toDouble(): Double = (+(this.asDynamic())).unsafeCast<Double>().also {
      if (it.isNaN() && !this.isNaN() || it == 0.0 && this.isBlank())
          numberFormatError(this)
  }
  ```
  Tests for those values are ignored for JS platform.

@sschuberth please, take a look at the PR when you have time. Thank you!

### DIFF
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
@@ -20,7 +20,6 @@ package com.charleskorn.kaml
 
 import io.kotest.assertions.asClue
 import io.kotest.assertions.throwables.shouldThrow
-import io.kotest.assertions.withClue
 import io.kotest.common.Platform
 import io.kotest.common.platform
 import io.kotest.core.test.Enabled
--- a/src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
+++ b/src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
@@ -25,6 +25,8 @@ import io.kotest.common.Platform
 import io.kotest.common.platform
 import io.kotest.core.test.Enabled
 import io.kotest.core.test.EnabledOrReasonIf
+import io.kotest.matchers.doubles.shouldBeNaN
+import io.kotest.matchers.floats.shouldBeNaN
 import io.kotest.matchers.shouldBe
 
 class YamlScalarTest : FlatFunSpec({
@@ -180,12 +182,9 @@ class YamlScalarTest : FlatFunSpec({
 
                         test("converts it to the expected double") {
                             if (expectedResult.isNaN()) {
-                                // this is not correct to compare NaNs
-                                // instead isNaN method should be invoked
-                                // to check whether the value is NaN or not
-                                withClue("$result should be NaN") {
-                                    result.isNaN() shouldBe true
-                                }
+                                // comparing NaNs requires special code
+                                // as they must not be compared via == / equals()
+                                result.shouldBeNaN()
                             } else {
                                 result shouldBe expectedResult
                             }
@@ -225,12 +224,9 @@ class YamlScalarTest : FlatFunSpec({
 
                         test("converts it to the expected float") {
                             if (expectedResult.isNaN()) {
-                                // this is not correct to compare NaNs
-                                // instead isNaN method should be invoked
-                                // to check whether the value is NaN or not
-                                withClue("$result should be NaN") {
-                                    result.isNaN() shouldBe true
-                                }
+                                // comparing NaNs requires special code
+                                // as they must not be compared via == / equals()
+                                result.shouldBeNaN()
                             } else {
                                 result shouldBe expectedResult
                             }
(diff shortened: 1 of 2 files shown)


src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
```diff
@@ -174,7 +179,16 @@ class YamlScalarTest : FlatFunSpec({
                         val result = scalar.toDouble()
 
                         test("converts it to the expected double") {
-                            result shouldBe expectedResult
+                            if (expectedResult.isNaN()) {
+                                // this is not correct to compare NaNs
+                                // instead isNaN method should be invoked
+                                // to check whether the value is NaN or not
```
sschuberth, line 185: I propose to reword this like:

> Comparing NaNs requires special code as they must not be compared via `==` / `equals()`.

src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
```diff
@@ -174,7 +179,16 @@ class YamlScalarTest : FlatFunSpec({
...
+                                withClue("$result should be NaN") {
+                                    result.isNaN() shouldBe true
+                                }
```
sschuberth, line 188: Can we instead use the dedicated `shouldBeNaN` matcher?

src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
```diff
@@ -210,7 +224,16 @@ class YamlScalarTest : FlatFunSpec({
                         val result = scalar.toFloat()
 
                         test("converts it to the expected float") {
-                            result shouldBe expectedResult
+                            if (expectedResult.isNaN()) {
```
sschuberth, line 227: Same comments as above apply to this hunk.

src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
```diff
@@ -228,36 +251,45 @@ class YamlScalarTest : FlatFunSpec({
                 "+",
                 "",
             ).forEach { content ->
+                val ignoreValidFloatingPointsForJs: EnabledOrReasonIf = {
+                    if (platform == Platform.JS && content in setOf("0x2", "0o2")) {
```
sschuberth, line 255: Is "0o2" really correct? Looks like a typo.

src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
```diff
@@ -228,36 +251,45 @@ class YamlScalarTest : FlatFunSpec({
...
+                        Enabled.disabled("$content is a valid floating value for JS due to dynamic cast")
+                    } else {
+                        Enabled.enabled
+                    }
+                }
```
sschuberth, line 260: Nit: Empty line here for visual separation of blocks?

sschuberth:
> @sschuberth please, take a look at the PR when you have time.

Done, though I'm not sure why, as I have no approval rights on this repo 😉

OptimumCode:
Thank you for your review @sschuberth!
I wanted to tag @charleskorn but he does not appear in the completion list for some reason, and I accidentally tagged you instead)) don't know why - probably because it is a bit late here already)

src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
OptimumCode: I didn't know there was a function like this. Thanks!

src/commonTest/kotlin/com/charleskorn/kaml/YamlScalarTest.kt
OptimumCode: This version sounds better, thanks

sschuberth:
> I accidentally tagged you instead

No worries, I hope my review was useful still 😸

OptimumCode:
> No worries, I hope my review was useful still 😸
It definitely was! Thanks again

charleskorn:
Thanks @OptimumCode and @sschuberth!

PR #587, opened 2024-07-08
Encoding and Decoding case insensitive enums

AbhigyaKrishna:
Implemented case-insensitive enumeration encoding and decoding.

# Changelogs
* [x] Configuration value for `decodeEnumCaseInsensitive` defaults to `false`.
* [x] Implementation of lenient deserialization in enum decoder.
* [x] Emits lowercase enum names when configured to `true` for consistency with `Json` format.
* [x] Implemented tests for case insensitive enums deserialization.

```kt
enum class TestEnum {
    VALUE
}

val yaml = Yaml(configuration = YamlConfiguration(decodeEnumCaseInsensitive = true))

val inResult = yaml.encodeToString(TestEnum.VALUE) // Outputs: value
val outResult = yaml.decodeFromString<TestEnum>("value") // Outputs: TestEnum.VALUE
```


rasros:
I need this as well. We switched from json files to yaml but are missing case-insensitive enum deserialization that the kotlinx serialization library has. Would love to see this merged soon if possible!

src/commonMain/kotlin/com/charleskorn/kaml/YamlScalarInput.kt
```diff
@@ -42,13 +42,22 @@ internal class YamlScalarInput(val scalar: YamlScalar, yaml: Yaml, context: Seri
...
-            .joinToString(", ")
+
+        if (configuration.decodeEnumCaseInsensitive) {
+            val lowerChoices = choices.map { it.lowercase() }
+            val lowerContent = scalar.content.lowercase()
+
+            val idx = lowerChoices.indexOf(lowerContent)
+            if (idx != -1) {
+                return idx
+            }
```
OptimumCode, line 56: suggestion: You can simplify this part. Instead of create a lowercase version of each element name you can use `equals(ignoreCase = true)` to find the element you need
```suggestion
            val idx = choices.indexOfFirst { it.equals(scalar.content, ignoreCase = true) }
            if (idx != -1) {
                return idx
            }
```

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -93,7 +93,15 @@ internal class YamlOutput(
         }
     }
 
-    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) = emitQuotedScalar(enumDescriptor.getElementName(index), configuration.singleLineStringStyle.scalarStyle)
+    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) {
+        val element = if (configuration.decodeEnumCaseInsensitive) { // use lower case enums
+            enumDescriptor.getElementName(index).lowercase()
```
OptimumCode, line 98: note: The configuration property states that it is related to decoding. Probably, the encoding should not be changed at all

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -52,6 +54,8 @@ public data class YamlConfiguration(
     internal val allowAnchorsAndAliases: Boolean = false,
     internal val yamlNamingStrategy: YamlNamingStrategy? = null,
     internal val codePointLimit: Int? = null,
+    @ExperimentalSerializationApi
```
OptimumCode, line 57: note: From my understanding, the idea of experimental annotations is that each project has its own annotations that can be used to mark experimental parts of the API. If this part of the API is experimental then, probably, you should create an experimental annotation in this repo and use in instead of the one declared in kotlinx-serialization library

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
AbhigyaKrishna: This is done to make the encoding consistent with how json handles it.

OptimumCode: By "json", do you mean kotlinx-serialization-json library?

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
AbhigyaKrishna: As of the `kaml` project is a extension of the `kotlinx.serialization` project. It would be more appropriate to use their experimental annotations. It will also make the code consistent across other kotlinx serializers.

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
AbhigyaKrishna: Yes

OptimumCode: I might be missing something, but I don't see a conversion to lower-case in kotlinx-serialization-json
- [stream encoding](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt#L228)
- [tree encoding](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/internal/TreeJsonEncoder.kt#L104)

Could you please point out where this is handled?

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
OptimumCode: Got it. Thank you for clarification. Also checked some other `kotlinx.serialization` libraries - they use the same approach. For example, [ktoml](https://github.com/orchestr7/ktoml/blob/main/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlArrayDecoder.kt#L20)

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
AbhigyaKrishna: Sorry, I misinterpreted their [names matching strategy](https://github.com/Kotlin/kotlinx.serialization/blob/8723076c60baf11ef151c6a267e208f36024eb8b/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonNamesMap.kt#L85) and [schema cache](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/internal/SchemaCache.kt) with lowercase encoding of enums.

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -93,7 +93,11 @@ internal class YamlOutput(
         }
     }
 
-    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) = emitQuotedScalar(enumDescriptor.getElementName(index), configuration.singleLineStringStyle.scalarStyle)
+    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) {
+        val element = enumDescriptor.getElementName(index)
+
+        emitQuotedScalar(element, configuration.singleLineStringStyle.scalarStyle)
+    }
```
OptimumCode, line 100: note: I think if we don't change the encoding we should revert the code to the original state

AbhigyaKrishna: Its the same code, rather descriptive.

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -93,7 +93,11 @@ internal class YamlOutput(
         }
     }
 
-    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) = emitQuotedScalar(enumDescriptor.getElementName(index), configuration.singleLineStringStyle.scalarStyle)
```
charleskorn, line 96: Could you please revert this change? It's not relevant to this PR.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
charleskorn, line 57: I don't think there's a need to mark this as experimental.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -52,6 +54,8 @@ public data class YamlConfiguration(
...
+    internal val decodeEnumCaseInsensitive: Boolean = false,
```
charleskorn, line 58: Could you please add an explanation of this property to the doc comment above?

charleskorn, line 58: I don't love this name... how about something like `enableCaseInsensitiveEnumDeserialization`?

charleskorn (commented):
Thanks for the PR!

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
OptimumCode: Hi, @charleskorn. The same naming is used in kotlinx-serialization-json

https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt#L36

Also, the current name is short and in the same time - self-descriptive. From my point of view, it would be better to have a name like this rather than something longer (and if we keep the experimental annotation the parameter can be renamed at some point)

If you still would prefer to change the name of the parameter maybe use "Decoding" instead of "Deserialization": `enableCaseInsensitiveEnumDecoding` (it makes the name a bit shorter🙂)

charleskorn: If that's the name that kotlinx.serialization uses, then staying consistent with that makes sense to me.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
AbhigyaKrishna: All the new api features are marked as experimental in the `kotlinx.serialization` library when they are introduced. As a matter of fact, its their convention.

charleskorn (approved):
Thanks again for the PR!

PR #591, opened 2024-07-17
Use Kotlinx.IO

mani1232:


charleskorn:
Could you please provide some more context on why you think this change should be made?

mani1232:
> Could you please provide some more context on why you think this change should be made?

It's not a very serious reason. But it can be done, as I don't see the need for Okio and its dependencies anymore.

JakeWharton:
Okio has no dependencies.

mani1232:
> Okio has no dependencies.

ok, but

kotlinx.io use 2.x.x kotlin version
The size of kotlinx.io is x2-2.5 times smaller

So far, other than alpha status, I don't see any downsides.

JakeWharton:
The version of Kotlin is irrelevant—it's a compiled dependency. The version a dependency uses does not affect its consumers other than setting a stdlib version floor.

Size also shouldn't matter. On every platform you only pay for what you use. The design of kotlinx.io is exactly the same as Okio so in the final binary each will cost the same.

The only consideration here is the fact that one API is ABI stable and the other is not. Should you use kotlinx.io over Okio once it's ABI stable? Always. Should you today when you will have to chase compatibility with every release? That's the only important question here.

mani1232:
> The version of Kotlin is irrelevant—it's a compiled dependency. The version a dependency uses does not affect its consumers other than setting a stdlib version floor.
> 
> Size also shouldn't matter. On every platform you only pay for what you use. The design of kotlinx.io is exactly the same as Okio so in the final binary each will cost the same.
> 
> The only consideration here is the fact that one API is ABI stable and the other is not. Should you use kotlinx.io over Okio once it's ABI stable? Always. Should you today when you will have to chase compatibility with every release? That's the only important question here.

about size - compared only jvm target

About the api status, yes, I agree.

charleskorn:
Sorry, until kotlinx.io's interface is stable, I'm not prepared to accept this change.

PR #596, opened 2024-08-03
feat(JVM): Annotations to set the ScalarStyle on a single field.

nielsbasjes:
A first draft of what I asked for in https://github.com/charleskorn/kaml/discussions/594
The main problem I faced is that the `fun encodeString(value: String)` as defined by Kotlin does not pass the descriptor and index.

So this first attempt works (on my machine) but does make that `this will not work in a multithreaded situation` and multiple yamls are written at the same time by the same instance.

So either that problem can be fixed in a different way or it is essentially impossible to build.

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
@@ -32,3 +32,27 @@ import kotlinx.serialization.SerialInfo
 public annotation class YamlComment(
     vararg val lines: String,
 )
+
+/**
+ * Write a String value if it is a single line in the specified ScalarStyle.
+ * This overrides the value specified in the [YamlConfiguration].
+ */
+@OptIn(ExperimentalSerializationApi::class)
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+@SerialInfo
+public annotation class YamlSingleLineStringStyle(
+    val singleLineStringStyle: SingleLineStringStyle,
+)
+
+/**
+ * Write a String value if it is a multiline in the specified ScalarStyle.
+ * This overrides the value specified in the [YamlConfiguration].
+ */
+@OptIn(ExperimentalSerializationApi::class)
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+@SerialInfo
+public annotation class YamlMultiLineStringStyle(
+    val multiLineStringStyle: MultiLineStringStyle,
+)
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
@@ -21,7 +21,7 @@ package com.charleskorn.kaml
 /**
  * Configuration options for parsing YAML to objects and serialising objects to YAML.
  *
- * * [encodeDefaults]: set to false to not write default property values to YAML (defaults to `true`)
+ * * [encodeDefaults]: set to `false` to not write default property values to YAML (defaults to `true`)
  * * [strictMode]: set to true to throw an exception when reading an object that has an unknown property, or false to ignore unknown properties (defaults to `true`)
  * * [extensionDefinitionPrefix]: prefix used on root-level keys (where document root is an object) to define extensions that can later be merged (defaults to `null`, which disables extensions altogether). See https://batect.dev/docs/reference/config#anchors-aliases-extensions-and-merging for example.
  * * [polymorphismStyle]: how to read or write the type of a polymorphic object:
@@ -31,9 +31,12 @@ package com.charleskorn.kaml
  * * [encodingIndentationSize]: number of spaces to use as indentation when encoding objects as YAML
  * * [breakScalarsAt]: maximum length of scalars when encoding objects as YAML (scalars exceeding this length will be split into multiple lines)
  * * [sequenceStyle]: how sequences (aka lists and arrays) should be formatted. See [SequenceStyle] for an example of each
+ * * [singleLineStringStyle]: the style in which a single line String value is written. Can be overruled for a specific field with the [YamlSingleLineStringStyle] annotation.
+ * * [multiLineStringStyle]: the style in which a multi line String value is written. Can be overruled for a specific field with the [YamlMultiLineStringStyle] annotation.
  * * [ambiguousQuoteStyle]: how strings should be escaped when [singleLineStringStyle] is [SingleLineStringStyle.PlainExceptAmbiguous] and the value is ambiguous
  * * [sequenceBlockIndent]: number of spaces to use as indentation for sequences, if [sequenceStyle] set to [SequenceStyle.Block]
  * * [allowAnchorsAndAliases]: set to true to allow anchors and aliases when decoding YAML (defaults to `false`)
+ * * [yamlNamingStrategy]: The system that converts the field names in to the names used in the Yaml.
  * * [codePointLimit]: the maximum amount of code points allowed in the input YAML document (defaults to 3 MB)
  */
 public data class YamlConfiguration(
@@ -82,6 +85,7 @@ public enum class SequenceStyle {
 
 public enum class MultiLineStringStyle {
     Literal,
+    Folded,
     DoubleQuoted,
     SingleQuoted,
     Plain,
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
@@ -80,15 +80,24 @@ internal class YamlOutput(
     override fun encodeInt(value: Int) = emitPlainScalar(value.toString())
     override fun encodeLong(value: Long) = emitPlainScalar(value.toString())
     override fun encodeShort(value: Short) = emitPlainScalar(value.toString())
+
+    private var forcedSingleLineScalarStyle: SingleLineStringStyle? = null
+    private var forcedMultiLineScalarStyle:  MultiLineStringStyle? = null
+
     override fun encodeString(value: String) {
         if (shouldReadTypeName) {
             currentTypeName = value
             shouldReadTypeName = false
         } else {
+            val singleLineScalarStyle = forcedSingleLineScalarStyle ?.scalarStyle ?: configuration.singleLineStringStyle.scalarStyle
+            val multiLineScalarStyle  = forcedMultiLineScalarStyle  ?.scalarStyle ?: configuration.multiLineStringStyle.scalarStyle
             when {
-                value.contains('\n') -> emitQuotedScalar(value, configuration.multiLineStringStyle.scalarStyle)
-                configuration.singleLineStringStyle == SingleLineStringStyle.PlainExceptAmbiguous && value.isAmbiguous() -> emitQuotedScalar(value, configuration.ambiguousQuoteStyle.scalarStyle)
-                else -> emitQuotedScalar(value, configuration.singleLineStringStyle.scalarStyle)
+                value.contains('\n')
+                    -> emitScalar(value, multiLineScalarStyle)
+                configuration.singleLineStringStyle == SingleLineStringStyle.PlainExceptAmbiguous && value.isAmbiguous()
+                    -> emitQuotedScalar(value, configuration.ambiguousQuoteStyle.scalarStyle)
+                else
+                    -> emitScalar(value, singleLineScalarStyle)
             }
         }
     }
@@ -98,6 +107,13 @@ internal class YamlOutput(
     private fun emitPlainScalar(value: String) = emitScalar(value, ScalarStyle.PLAIN)
     private fun emitQuotedScalar(value: String, scalarStyle: ScalarStyle) = emitScalar(value, scalarStyle)
 
+
+    private inline fun <reified R> SerialDescriptor.getAnnotation(index: Int): R? {
+        return getElementAnnotations(index)
+            .filterIsInstance<R>()
+            .firstOrNull()
+    }
+
     override fun encodeElement(descriptor: SerialDescriptor, index: Int): Boolean {
         encodeComment(descriptor, index)
 
@@ -106,6 +122,10 @@ internal class YamlOutput(
             val serializedName = configuration.yamlNamingStrategy?.serialNameForYaml(elementName) ?: elementName
             emitPlainScalar(serializedName)
         }
+
+        // If this field was annotated we overrule the used ScalarStyle with the annotation
+        forcedSingleLineScalarStyle = descriptor.getAnnotation<YamlSingleLineStringStyle>(index)?.singleLineStringStyle
+        forcedMultiLineScalarStyle = descriptor.getAnnotation<YamlMultiLineStringStyle>(index)?.multiLineStringStyle
 
         return super.encodeElement(descriptor, index)
     }
@@ -159,9 +179,7 @@ internal class YamlOutput(
     }
 
     private fun encodeComment(descriptor: SerialDescriptor, index: Int) {
-        val commentAnno = descriptor.getElementAnnotations(index)
-            .filterIsInstance<YamlComment>()
-            .firstOrNull() ?: return
+        val commentAnno = descriptor.getAnnotation<YamlComment>(index) ?: return
 
         for (line in commentAnno.lines) {
             emitter.emit(CommentEvent(CommentType.BLOCK, " $line", null, null))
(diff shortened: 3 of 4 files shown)


src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -80,15 +80,25 @@ internal class YamlOutput(
     override fun encodeInt(value: Int) = emitPlainScalar(value.toString())
     override fun encodeLong(value: Long) = emitPlainScalar(value.toString())
     override fun encodeShort(value: Short) = emitPlainScalar(value.toString())
+
+    // FIXME: This will go really bad in mutlithreaded situations
```
charleskorn, line 84: Multi threaded serialization is not possible, so this is fine.

src/commonMain/kotlin/com/charleskorn/kaml/YamlAnnotations.kt
```diff
@@ -32,3 +32,32 @@ import kotlinx.serialization.SerialInfo
...
+
+
+/**
+ * If specified the String produced by the serializing will be written as a scalar using the 'Plain' style.
+ */
+@OptIn(ExperimentalSerializationApi::class)
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+@SerialInfo
+public annotation class YamlWriteAsPlainScalar
```
charleskorn, line 63: Rather than multiple annotations, how about we have just one, with an enum specifying the style? Then we don't have to worry about multiple different conflicting annotations being applied to the same property.

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
```diff
@@ -190,3 +221,16 @@ sealed interface Animal {
     @Serializable
     data class Cat(val name: String) : Animal
 }
+
+// ------------------------------------------
```
charleskorn, line 225: This comment is not necessary.

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
```diff
@@ -176,6 +176,37 @@ class JvmYamlWritingTest : DescribeSpec({
...
+                "\n" +
+                "  Folded 3\n" +
+                "plain: 'Plain 1\n" +
+                "\n" +
+                "  Plain 2\n" +
+                "\n" +
+                "  Plain 3\n" +
+                "\n" +
+                "  '\n"
+        }
```
charleskorn, line 209: Could you please use a raw string here? It'll be much clearer.

src/commonMain/kotlin/com/charleskorn/kaml/YamlAnnotations.kt
```diff
@@ -32,3 +32,32 @@ import kotlinx.serialization.SerialInfo
 public annotation class YamlComment(
     vararg val lines: String,
```
charleskorn, line 33: This file could just become `Annotations.kt`.

charleskorn (commented):
Thanks for the PR, the general approach looks good to me. 

Could you please also add a comment to the relevant properties on `YamlConfiguration` to explain that these annotations take precedence over any configuration there?

src/commonMain/kotlin/com/charleskorn/kaml/YamlAnnotations.kt
nielsbasjes: Yes, indeed much better.

nielsbasjes:
I have updated the documentation a bit more.
Now in addition to the YamlConfiguration being shown as this:
image

You now also get this:

image

nielsbasjes:
I have handled your review comments and it now has 2 annotations YamlWriteSingleLineStringUsingScalarStyle and YamlWriteMultiLineStringUsingScalarStyle.

I have also added a few more tests: different combinations of these 2 and having a single or multiline value)

I did not (yet) add a similar annotation for the ambiguous case because I did not yet understand how that works and should work.

src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
```diff
@@ -32,3 +33,27 @@ import kotlinx.serialization.SerialInfo
...
+/**
+ * Write a String value if it is a single line in the specified ScalarStyle.
+ * This overrides the value specified in the [YamlConfiguration].
+ */
+@OptIn(ExperimentalSerializationApi::class)
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+@SerialInfo
+public annotation class YamlWriteSingleLineStringUsingScalarStyle(
+    val scalarStyle: ScalarStyle,
```
OptimumCode, line 46: issue: the SnakeKMP dependency is specified as `implementation`. This means the end user won't be able to use `ScalarStyle` without explicitly adding the SnakeKMP into their dependencies. We probably should create a new enumeration in this library that the end user could use.
I am not sure how aliases work but if we can create an alias for `ScalarStyle` and user would be able to import it without having SnakeKMP explicitly specified in their dependency, that could be an alternative to the suggestion above.

nielsbasjes: When I look at the YamlOutput code I find this `ScalarStyle` class to be used all over the place. How can a consumer of this library use it without using this?

OptimumCode: Very easy: usage is encapsulated inside YamlOutput class. You can say it is implementation details. The end user doesn't have to know about them at all.

This is all about compile and runtime dependencies. The first group is available during compilation (and usually in runtime). The second - during execution when you run the program.
The SnakeKMP is both compile and runtime dependency for this library. But it will be only a runtime dependency for those who consumes kaml. Meaning, it won't be available at compilation time. And because of that, the end user won't be able to use this class in their code - it won't simply compile.

This [link](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies) contains additional information about api and implementation dependencies

nielsbasjes: So in cases where a class is annotated with this but the serialization using Kaml is not actually not executed: an application can exist without the SnakeKMP dependency.
Ok, I find this a rare case but in theory this can exist.

I'll have a look at fixing this.

OptimumCode: Not really, the SnakeKMP is a runtime dependency for the end user. The library consumer will have it in their **runtime** classpath. But in order to use the ScalarStyle class in their code the user must have it in **compile** classpath.
Sorry, if my explanation doesn't make it clear. I hope you could find some time to take a look at the link I shared in the previous comment - it should make everything clear.

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -81,16 +81,25 @@ internal class YamlOutput(
     override fun encodeLong(value: Long) = emitPlainScalar(value.toString())
     override fun encodeShort(value: Short) = emitPlainScalar(value.toString())
 
-    private var forcedSingleLineScalarStyle: ScalarStyle? = null
-    private var forcedMultiLineScalarStyle: ScalarStyle? = null
+    private var forcedSingleLineScalarStyle: StringScalarStyle? = null
+    private var forcedMultiLineScalarStyle:  StringScalarStyle? = null
+
+    private fun StringScalarStyle.toScalarStyle() =
```
OptimumCode, line 87: suggestion: I am not sure if it is needed at all but it would be probably good to know if there is a `ScalarStyle` value that is not covered by the `StringScalarStyle` enum (in case the new value is added in one of the versions).

To do that we probably should create a test that iterates over `StringScalarStyle`'s values, converts them to `ScalarStyle` and compares the resulted set with all values available in `ScalarStyle` and reports the difference

nielsbasjes: If that happens the code will fail to build:
image

So no test is needed to ensure this.

nielsbasjes: I double checked this is not only an IDEA error; also the compile fails over this.

OptimumCode: Hi, @nielsbasjes. This compile error only happens if you have a new value added to `StringScalarStyle`. However, if a new value is added to `ScalarStyle` the code will still compile (because the `when` expression will still be exhausted). As I said, I am not sure if this test is required (YAML spec does not have other types of scalar as far as I know) but still it could be good to have a test to cover this.

Also, nobody is safe from a bug like this (but this is another problem...):
```kotlin
return when (this) {
    StringScalarStyle.DOUBLE_QUOTED -> ScalarStyle.DOUBLE_QUOTED
    StringScalarStyle.SINGLE_QUOTED -> ScalarStyle.SINGLE_QUOTED
    StringScalarStyle.LITERAL       -> ScalarStyle.LITERAL
    StringScalarStyle.FOLDED        -> ScalarStyle.LITERAL // duplicate
    StringScalarStyle.PLAIN         -> ScalarStyle.PLAIN
}
```

nielsbasjes: Test added.

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
```diff
@@ -282,6 +283,15 @@ class JvmYamlWritingTest : DescribeSpec({
                 """.trimIndent()
         }
 
+        describe("Ensure all StringScalarStyle and ScalarStyle values match 1 on 1") {
+            val stringScalarStyles = StringScalarStyle.entries.map { it.name }
+            val scalarStyles = ScalarStyle.entries.map { it.name }
+            stringScalarStyles.shouldContainExactlyInAnyOrder(scalarStyles)
```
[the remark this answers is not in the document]
nielsbasjes: Fixed

src/commonMain/kotlin/com/charleskorn/kaml/YamlOutput.kt
```diff
@@ -259,3 +250,12 @@ internal class YamlOutput(
         private val ALL_EXPLICIT = ImplicitTuple(false, false)
     }
 }
+
+public fun StringScalarStyle.toScalarStyle(): ScalarStyle =
```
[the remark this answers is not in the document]
nielsbasjes: Thanks, I didn't know about the `internal` keyword yet. Awesome improvement over Java.

nielsbasjes:
NOTE: There is no documentation in YamlConfiguration for `yamlNamingStrategy`. 
If you have a line I can add it right away.

nielsbasjes:
I've handled the feedback as best as I understand what you want to see.
Looking forward to any other changes you would like me to make.

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -31,26 +31,46 @@ package com.charleskorn.kaml
...
     internal val multiLineStringStyle: MultiLineStringStyle = singleLineStringStyle.multiLineStringStyle,
+    /** How strings should be escaped when [singleLineStringStyle] is [SingleLineStringStyle.PlainExceptAmbiguous] and the value is ambiguous */
     internal val ambiguousQuoteStyle: AmbiguousQuoteStyle = AmbiguousQuoteStyle.DoubleQuoted,
+    /** Number of spaces to use as indentation for sequences, if [sequenceStyle] set to [SequenceStyle.Block] */
     internal val sequenceBlockIndent: Int = 0,
+    /** Set to true to allow anchors and aliases when decoding YAML (defaults to `false`) */
     internal val allowAnchorsAndAliases: Boolean = false,
     internal val yamlNamingStrategy: YamlNamingStrategy? = null,
+    /** The maximum amount of code points allowed in the input YAML document (defaults to 3 MB) */
     internal val codePointLimit: Int? = null,
```
charleskorn, line 74: There's no need to duplicate the documentation like this - the comment above is sufficient.

src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
```diff
@@ -32,3 +32,38 @@ import kotlinx.serialization.SerialInfo
...
+
+/**
+ * Write a String value if it is a single line in the specified ScalarStyle.
+ * This overrides the value specified in the [YamlConfiguration].
+ */
+@OptIn(ExperimentalSerializationApi::class)
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+@SerialInfo
+public annotation class YamlWriteSingleLineStringUsingScalarStyle(
```
charleskorn, line 44: Perhaps `YamlSingleLineStringStyle` would be enough? This would also keep the symmetry with `YamlConfiguration`

(Similar comment applies for the multi-line annotation below)

src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
```diff
@@ -32,3 +32,38 @@ import kotlinx.serialization.SerialInfo
...
+/**
+ * Write a String value if it is a multiline in the specified ScalarStyle.
+ * This overrides the value specified in the [YamlConfiguration].
+ */
+@OptIn(ExperimentalSerializationApi::class)
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+@SerialInfo
+public annotation class YamlWriteMultiLineStringUsingScalarStyle(
+    val stringScalarStyle: StringScalarStyle,
```
charleskorn, line 57: Any reason not to use the same enum `YamlConfiguration` uses?

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
```diff
@@ -176,6 +178,121 @@ class JvmYamlWritingTest : DescribeSpec({
...
+                    
+                      Plain 3
+                    
+                      '
+                
+                """.trimIndent()
+        }
+
+        describe("Ensure all StringScalarStyle and ScalarStyle values match 1 on 1") {
+            // Mapping the StringScalarStyles to ScalarStyle must result in 100% coverage
```
charleskorn, line 287: I don't think this test is necessary. There are a fixed set of styles in YAML, and we have support for all of them already.

charleskorn (commented):
Thanks for the PR!

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
nielsbasjes: Ok, I'll remove this addition of mine.

src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
nielsbasjes: Yes, sounds good.

src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
nielsbasjes: Yes, that seems like a good option. This evolved from a different initial proposal.

I would propose then to add the `Folded` to the `MultiLineStringStyle` to make it complete.

nielsbasjes: Should I add the Folded here in this MR or make it a separate thing?

charleskorn: Adding it as part of this PR is fine.

nielsbasjes: FYI: I do see that the `Literal` and `Folded` formats are not possible for SingleLine cases (which is fine for me and is the same as the current situation).

src/jvmTest/kotlin/com/charleskorn/kaml/JvmYamlWritingTest.kt
nielsbasjes: Removed

src/commonMain/kotlin/com/charleskorn/kaml/Annotations.kt
nielsbasjes: Using existing enums, added Folded

nielsbasjes:
Squashed all commits I made to make it a lot cleaner

charleskorn (commented):
This looks good to me. Could you please fix the linting failures?

charleskorn:
Thanks @nielsbasjes!

PR #607, opened 2024-09-05
feat: add `YamlContentPolymorphicSerializer`, similar to `JsonContentPolymorphicSerializer`

Jojo4GH:
- [x] Implemented `YamlContentPolymorphicSerializer` in the same manner as [`JsonContentPolymorphicSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-content-polymorphic-serializer/)
- [x] Implemented tests for `YamlContentPolymorphicSerializer`
- Depends on:
  - https://github.com/charleskorn/kaml/pull/617 
  - https://github.com/charleskorn/kaml/pull/618

Example usage:
```kotlin
@Serializable(with = AnimalSerializer::class)
sealed interface Animal {
    @Serializable data class Cat(val isPurring: Boolean)
    @Serializable data class Dog(val isBarking: Boolean)
}

object AnimalSerializer : YamlContentPolymorphicSerializer<Animal>(Animal::class) {
    override fun selectDeserializer(node: YamlNode): DeserializationStrategy<Animal> = when {
        node.yamlMap.getKey("isPurring") != null -> Animal.Cat.serializer()
        node.yamlMap.getKey("isBarking") != null -> Animal.Dog.serializer()
        else -> error("Unknown animal")
    }
}
```


Jojo4GH:
Optional improvement: Make the `YamlNode` API a bit more consumer friendly (especially for getting keys in `YamlMap`)

If this should be implemented in this PR, then I can do it

charleskorn:
> * [x]  Fixed bug with `YamlInput::node`
> * [x]  Fixed bug with sealed polymorphism for scalars represented as inline value classes

Could you please elaborate on these? Perhaps they could be separate PRs to help keep this one smaller and easier to review.



> Optional improvement: Make the `YamlNode` API a bit more consumer friendly (especially for getting keys in `YamlMap`)
> 
> If this should be implemented in this PR, then I can do it

Sounds like something for a separate PR.

Jojo4GH:
> Fixed bug with `YamlInput::node`

Previously the `node` property on the decoders was not traversed in deep YAML trees (e.g. with `YamlMapLikeInputBase` and `YamlListInput`). This behaviour is different from the JSON serialization library and makes it impossible to implement a content based polymorphic `KSerializer` in a meaningful way without knowing of the whole YAML tree. See also the `DecodingFromYamlNodeSerializer` in `YamlReadingTest.kt` for an affected case

> Fixed bug with sealed polymorphism for scalars represented as inline value classes

Previously it was not possible to implement a content based polymorphic serializer in the case that one of the types involved is a scalar. Different from the JSON serialization library it was therefore impossible to implement something like:

```yaml
myCats:
  myFirstCat: "Bella"
  mySecondCat:
    name: "Daisy"
    age: 2
```

```kotlin
typealias MyCats = Map<String, CatDetails>

@Serializable(with = MyContentBasedSerializer::class)
sealed interface CatDetails {

    @Serializable
    value class InlineName(val name: String) : CatDetails

    @Serializable
    data class More(
        val name: String,
        val age: String?
    ) : CatDetails
}

object MyContentBasedSerializer : YamlContentPolymorphicSerializer<CatDetails>(CatDetails::class) {
    override fun selectDeserializer(node: YamlNode) = when (node) {
        is YamlScalar -> CatDetails.InlineName.serializer()
        is YamlMap -> CatDetails.More.serializer()
        else -> error("Unsupported node type ${node::class.simpleName}")
    }
}
```

> Optional improvement: [...]

Will do it sometime later in a separate PR.

charleskorn:
Could you perhaps split the two bug fixes out as their own separate PR? This is quite a big PR and so breaking it down into smaller pieces will make it easier for me to review.

Jojo4GH:
> Could you please split the two bug fixes out as their own separate PR? This is quite a big PR and so breaking it down into smaller pieces will make it easier for me to review.

This PR now depends on:
- https://github.com/charleskorn/kaml/pull/617 
- https://github.com/charleskorn/kaml/pull/618

Jojo4GH:
> > Could you please split the two bug fixes out as their own separate PR? This is quite a big PR and so breaking it down into smaller pieces will make it easier for me to review.
> 
> This PR now depends on:
> 
> * [fix: `Decoder` does not represent the current node correctly #617](https://github.com/charleskorn/kaml/pull/617)
> * [Prepare fix for contextual deserialization of polymorphic scalars #618](https://github.com/charleskorn/kaml/pull/618)

The necessary PRs are now merged so I merged main for easier review

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
```diff
@@ -115,6 +128,8 @@ public sealed class YamlInput(
         private fun YamlMap.withoutKey(key: String): YamlMap {
             return this.copy(entries = entries.filterKeys { it.content != key })
         }
+
+        private val SerialDescriptor.isContentBasedPolymorphic get() = serialName.startsWith(YamlContentPolymorphicSerializer::class.simpleName!!)
```
charleskorn, line 132: Rather than checking the name like this, what if we introduced an annotation that is added to the `SerialDescriptor` created by `YamlContentPolymorphicSerializer`?

src/commonTest/kotlin/com/charleskorn/kaml/YamlContentPolymorphicSerializer.kt
charleskorn, line 1: Could you please rename this file to `YamlContentPolymorphicSerializerTest.kt`?

Jojo4GH: Will do 👍🏻

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
Jojo4GH: The annotation is implemented as `internal` for now. Or should it be `public`?

src/commonTest/kotlin/com/charleskorn/kaml/YamlContentPolymorphicSerializer.kt
Jojo4GH: Done

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
charleskorn: `internal` seems fine to me, it's an implementation detail.

charleskorn (approved):
Thanks @Jojo4GH!

PR #617, opened 2024-09-16
fix: `Decoder` does not represent the current node correctly

Jojo4GH:
First bug from https://github.com/charleskorn/kaml/pull/607

> Previously the `node` property on the decoders was not traversed in deep YAML trees (e.g. with `YamlMapLikeInputBase` and `YamlListInput`). This behaviour is different from the JSON serialization library and makes it impossible to implement a content based polymorphic `KSerializer` in a meaningful way without knowing of the whole YAML tree. See also the `DecodingFromYamlNodeSerializer` in `YamlReadingTest.kt` for an affected case

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlListInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlListInput.kt
@@ -18,6 +18,7 @@
 
 package com.charleskorn.kaml
 
+import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.descriptors.SerialDescriptor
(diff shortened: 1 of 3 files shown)


charleskorn (commented):
Could you please add a test that demonstrates the issue this PR is trying to fix?

Jojo4GH:
> Could you please add a test that demonstrates the issue this PR is trying to fix?

There is the change in YamlReadingTest that gets fixed by this PR. Should I add more?

charleskorn:
> > Could you please add a test that demonstrates the issue this PR is trying to fix?
> 
> There is the change in YamlReadingTest that gets fixed by this PR. Should I add more?

It's not clear to me how the two changes are related, so yes: please add a simple test case that shows the issue and how this fixes it.

charleskorn:
Thanks @Jojo4GH!

PR #618, opened 2024-09-16
Prepare fix for contextual deserialization of polymorphic scalars

Jojo4GH:
Second bug from https://github.com/charleskorn/kaml/pull/607 (as preparation for the PR)

> Previously it was not possible to implement a content based polymorphic serializer in the case that one of the types involved is a scalar. Different from the JSON serialization library it was therefore impossible to implement something like:

```yaml
myCats:
  myFirstCat: "Bella"
  mySecondCat:
    name: "Daisy"
    age: 2
```

```kotlin
typealias MyCats = Map<String, CatDetails>

@Serializable(with = MyContentBasedSerializer::class)
sealed interface CatDetails {

    @Serializable
    value class InlineName(val name: String) : CatDetails

    @Serializable
    data class More(
        val name: String,
        val age: String?
    ) : CatDetails
}

object MyContentBasedSerializer : YamlContentPolymorphicSerializer<CatDetails>(CatDetails::class) {
    override fun selectDeserializer(node: YamlNode) = when (node) {
        is YamlScalar -> CatDetails.InlineName.serializer()
        is YamlMap -> CatDetails.More.serializer()
        else -> error("Unsupported node type ${node::class.simpleName}")
    }
}
```

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlContextualInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlContextualInput.kt
@@ -24,7 +24,11 @@ import kotlinx.serialization.modules.SerializersModule
 
 internal class YamlContextualInput(node: YamlNode, yaml: Yaml, context: SerializersModule, configuration: YamlConfiguration) : YamlInput(node, yaml, context, configuration) {
     override fun decodeElementIndex(descriptor: SerialDescriptor): Int = throw IllegalStateException("Must call beginStructure() and use returned Decoder")
-    override fun decodeValue(): Any = throw IllegalStateException("Must call beginStructure() and use returned Decoder")
+    override fun decodeValue(): Any = when (node) {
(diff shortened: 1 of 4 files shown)


charleskorn (commented):
It's not clear to me why this change is necessary - why can't a contextual polymorphic serializer look at the value of `node`?

Jojo4GH:
> It's not clear to me why this change is necessary - why can't a contextual polymorphic serializer look at the value of `node`?

Since `YamlContentPolymorphicSerializer` uses `PolymorphicKind.SEALED` (just like the `JsonContentPolymorphicSerializer`) when selecting an existing serializer that expects a `YamlScalar` the `IllegalStateException("Must call beginStructure() and use returned Decoder")` gets thrown erroneously.  
Remember that `fun selectDeserializer(node: YamlNode): DeserializationStrategy<T>` returns not a value but a deserializer.

charleskorn (approved):

PR #620, opened 2024-09-23
feat: allow to set a maximum amount of aliases while parsing to protect against billion laughs attack and similar

Vampire:


Vampire:
In the current code of the PR the cumulated alias weight of all nodes so far is used as a criterion.
An alternative could be to stop parsing if an individual node exceeds a given weight limit.

charleskorn:
Is this necessary? `YamlConfiguration` already has a `allowAnchorsAndAliases` property that defaults to `false`. I think adding what is proposed here adds a lot of complexity and I worry that it provides a false sense of security.

Vampire:
Yes, this is definitely necessary. 

If you set `allowAnchorsAndAliases` to `false` you cannot use anchors and aliases.

If you need them and thus set it to `true`, you are one blink away from a DoS attack which is very trivial. 

With this protection, the DoS attack is prevented. It is a similar protection e.g. snakeyaml has built in if you use the high-level API (you use the low-level one), or also the YAML NPM package.

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -1305,6 +1305,99 @@ class YamlReadingTest : FlatFunSpec({
...
+
+                        exception.asClue {
+                            it.message shouldBe "Parsing anchors and aliases is disabled."
+                            it.line shouldBe 1
+                            it.column shouldBe 18
+                        }
+                    }
+                }
+
+                context("parsing anchors and aliases is enabled, max aliases count 1") {
```
charleskorn, line 1345: Could you please also add a test case with max alias count set to 2, to verify that the content is allowed in this case?

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -56,6 +57,7 @@ public data class YamlConfiguration(
     internal val ambiguousQuoteStyle: AmbiguousQuoteStyle = AmbiguousQuoteStyle.DoubleQuoted,
     internal val sequenceBlockIndent: Int = 0,
     internal val allowAnchorsAndAliases: Boolean = false,
+    internal val maxAliasCount: Int = 100,
```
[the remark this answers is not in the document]
Vampire: > Rather than introducing a maxAliasCount property, what if we introduced a sealed class and used that to replace allowAnchorsAndAliases?

Sure, I'm all-in.
Just thought I do it in a backward compatible way. :-)

> I'd also prefer if maxAliasCount was a nullable unsigned integer - null indicating that there is no limit to the number of aliases allowed.

Definitely. Probably glimpsed too much at the other libs how they did it. :-)

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeReader.kt
```diff
@@ -29,37 +29,43 @@ import it.krzeminski.snakeyaml.engine.kmp.events.SequenceStartEvent
 internal class YamlNodeReader(
     private val parser: YamlParser,
     private val extensionDefinitionPrefix: String? = null,
-    private val allowAnchorsAndAliases: Boolean = false,
+    private val maxAliasCount: Int = 0,
 ) {
     private val aliases = mutableMapOf<Anchor, YamlNode>()
+    private val nodeWeight = mutableMapOf<YamlNode, Int>()
```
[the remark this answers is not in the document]
Vampire: Puh, don't ask me such things more than 6 weeks after I wrote it. :-D
I have no idea whether I had a reason or not.
It might be an oversight, or a left-over of a previous implementation, I changed it quite a bit before I was satisfied.

src/commonMain/kotlin/com/charleskorn/kaml/Yaml.kt
```diff
@@ -80,8 +80,11 @@ public class Yaml(
 
     internal fun parseToYamlNode(source: Source): YamlNode {
         val parser = YamlParser(source, configuration.codePointLimit)
-        val reader =
-            YamlNodeReader(parser, configuration.extensionDefinitionPrefix, configuration.allowAnchorsAndAliases)
+        val reader = YamlNodeReader(
+            parser,
+            configuration.extensionDefinitionPrefix,
+            maxAliasCount = configuration.anchorsAndAliases.maxAliasCount,
```
charleskorn, line 86: The explicit parameter name isn't necessary here.

```suggestion
            configuration.anchorsAndAliases.maxAliasCount,
```

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -124,3 +124,16 @@ public enum class AmbiguousQuoteStyle {
...
+
+public sealed class AnchorsAndAliases {
+    internal abstract val maxAliasCount: UInt?
+
+    public data object Forbidden : AnchorsAndAliases() {
+        override val maxAliasCount: UInt = 0u
+    }
+
+    /**
+     * [maxAliasCount]: the maximum amount of aliases allowed in the input YAML document if allowed at all, `null` allows any amount (defaults to `100u`)
```
charleskorn, line 136:
```suggestion
     * [maxAliasCount]: the maximum amount of aliases allowed in the input YAML document if allowed at all, `null` allows any amount (defaults to `100`)
```

charleskorn (commented):
Two small suggestions, and then I think this is good to go!

src/commonMain/kotlin/com/charleskorn/kaml/YamlConfiguration.kt
```diff
@@ -37,7 +37,7 @@ import kotlinx.serialization.ExperimentalSerializationApi
  * * [multiLineStringStyle]: the style in which a multi line String value is written. Can be overruled for a specific field with the [YamlMultiLineStringStyle] annotation.
  * * [ambiguousQuoteStyle]: how strings should be escaped when [singleLineStringStyle] is [SingleLineStringStyle.PlainExceptAmbiguous] and the value is ambiguous
  * * [sequenceBlockIndent]: number of spaces to use as indentation for sequences, if [sequenceStyle] set to [SequenceStyle.Block]
- * * [allowAnchorsAndAliases]: set to true to allow anchors and aliases when decoding YAML (defaults to `false`)
+ * * [anchorsAndAliases]: set to [AnchorsAndAliases.Permitted] to allow anchors and aliases when decoding YAML (defaults to [AnchorsAndAliases.Forbidden]`)
```
charleskorn, line 40:
```suggestion
 * * [anchorsAndAliases]: set to [AnchorsAndAliases.Permitted] to allow anchors and aliases when decoding YAML (defaults to [AnchorsAndAliases.Forbidden])
```

charleskorn:
@Vampire could you please fix the linting failure?

Vampire:
Whoops, sorry, did the uglification now. :-)

charleskorn (approved):
Thanks @Vampire!

PR #642, opened 2024-11-17
feat: Make Yaml nodes serializable

### LINKED ISSUE #641: Make YamlNode serializable

I would like to deserialize structures, that are partly arbitrary. For example when I want to deserialize the [mustache specification], most of the structure can be properly deserialized with somethling like
```kotlin
@Serializable
data class Specification(
  val overview: String,
  val tests: List<Test>,
)

@Serializable
data class Test(
  val name: String,
  val desc: String,
  val template: String,
  val expected: String,
  val partials: Map<String, String> = emptyMap(),
  val data: YamlNode? = null, // not serializable
)
```
but the `data` can't be, because it can be arbitrary nested or even just a primitive value.

I would like to have `YamlNode` as `Serializable` (similar as done by the kotlinx team for the [`JsonElement`](https://github.com/Kotlin/kotlinx.serialization/blob/d4d066d72a9f92f06c640be5a36a22f75d0d7659/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt#L26)).

I could write it on my own in my project, but I think, this would be really valuable to the library. Another alternative is to read the whole structure as `YamlNode` and then parse everything on with the node, but that reduces the usability.

### Additional context

_No response_

EdwarDDay:
- fixes #641

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
@@ -62,6 +62,13 @@ public sealed class YamlInput(
             is YamlList -> when (descriptor.kind) {
                 is StructureKind.LIST -> YamlListInput(node, yaml, context, configuration)
                 is SerialKind.CONTEXTUAL -> createContextual(node, yaml, context, configuration, descriptor)
+                is PolymorphicKind -> {
+                    if (descriptor.isContentBasedPolymorphic) {
+                        createContextual(node, yaml, context, configuration, descriptor)
+                    } else {
+                        throw MissingTypeTagException(node.path)
+                    }
+                }
                 else -> throw IncorrectTypeException("Expected ${descriptor.kind.friendlyDescription}, but got a list", node.path)
             }
 
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
@@ -23,7 +23,6 @@ package com.charleskorn.kaml
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.InternalSerializationApi
 import kotlinx.serialization.KSerializer
-import kotlinx.serialization.SerializationException
 import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.builtins.MapSerializer
 import kotlinx.serialization.builtins.serializer
@@ -67,19 +66,10 @@ internal object YamlScalarSerializer : KSerializer<YamlScalar> {
 
     override fun serialize(encoder: Encoder, value: YamlScalar) {
         encoder.asYamlOutput()
-        try {
-            return encoder.encodeBoolean(value.toBoolean())
-        } catch (_: SerializationException) {
-        }
-        try {
-            return encoder.encodeLong(value.toLong())
-        } catch (_: SerializationException) {
-        }
-        try {
-            return encoder.encodeDouble(value.toDouble())
-        } catch (_: SerializationException) {
-        }
-        value.content.singleOrNull()?.also { return encoder.encodeChar(it) }
+        value.toBooleanOrNull()?.also { return encoder.encodeBoolean(it) }
+        value.toLongOrNull()?.also { return encoder.encodeLong(it) }
+        value.toDoubleOrNull()?.also { return encoder.encodeDouble(it) }
+        value.toCharOrNull()?.also { return encoder.encodeChar(it) }
         encoder.encodeString(value.content)
     }
 
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
@@ -20,7 +20,6 @@
 
 package com.charleskorn.kaml
 
-import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.InternalSerializationApi
 import kotlinx.serialization.KSerializer
@@ -39,8 +38,11 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 import kotlinx.serialization.encoding.encodeStructure
 
-internal object YamlNodeSerializer : YamlContentPolymorphicSerializer<YamlNode>(YamlNode::class) {
-    override val descriptor: SerialDescriptor = super.descriptor.nullable
+internal object YamlNodeSerializer : KSerializer<YamlNode> {
+    override val descriptor: SerialDescriptor =
+        buildSerialDescriptor("com.charleskorn.kaml.YamlNode", PolymorphicKind.SEALED) {
+            annotations += YamlContentPolymorphicSerializer.Marker()
+        }.nullable
 
     override fun serialize(encoder: Encoder, value: YamlNode) {
         encoder.asYamlOutput()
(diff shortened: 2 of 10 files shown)


src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
```diff
@@ -0,0 +1,173 @@
...
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.SerialKind
+import kotlinx.serialization.descriptors.buildSerialDescriptor
+import kotlinx.serialization.descriptors.nullable
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.encoding.encodeStructure
+
+internal object YamlNodeSerializer : YamlContentPolymorphicSerializer<YamlNode>(YamlNode::class) {
```
charleskorn, line 42: Can't this just be an ordinary `KSerializer`?

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
```diff
@@ -0,0 +1,173 @@
...
+        } catch (_: SerializationException) {
+        }
+        try {
+            return encoder.encodeLong(value.toLong())
+        } catch (_: SerializationException) {
+        }
+        try {
+            return encoder.encodeDouble(value.toDouble())
+        } catch (_: SerializationException) {
+        }
```
charleskorn, line 83: Rather than this chain of `try` / `catch`, couldn't we inspect the value and determine the best option to use?

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
```diff
@@ -62,6 +62,13 @@ public sealed class YamlInput(
             is YamlList -> when (descriptor.kind) {
                 is StructureKind.LIST -> YamlListInput(node, yaml, context, configuration)
                 is SerialKind.CONTEXTUAL -> createContextual(node, yaml, context, configuration, descriptor)
+                is PolymorphicKind -> {
+                    if (descriptor.isContentBasedPolymorphic) {
+                        createContextual(node, yaml, context, configuration, descriptor)
+                    } else {
+                        throw MissingTypeTagException(node.path)
+                    }
+                }
```
charleskorn, line 71: Why is this change required?

src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
```diff
@@ -729,6 +793,46 @@ class YamlWritingTest : FlatFunSpec({
...
+                      - 2
+                      - 3
+                """.trimIndent(),
+            ) as YamlMap
+            val expectedOutput = """
+                    text: "test"
+                    node:
+                      "'foo'": "'bar'"
+                      "'baz'": 1
+                      "'test'":
```
charleskorn, line 812: Why are the values quoted twice here?

charleskorn (commented):
Thanks for the PR @EdwarDDay!

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
EdwarDDay: Changed. I mainly did it, because the `Marker` annotation is needed for the polymorphic handling.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
EdwarDDay: I didn't want to copy the parsing code, to avoid duplicated code. Do you have an idea how to implement this without duplicating most of the `YamlScalar` code?
I mean `Boolean` has extra parsing code, integer values extra checks for octal, decimal and hexadecimal numbers, floating point numbers have extra checks for infinity and NaN. This was too much code for me to copy. Else the library gets inconsistent with future changes.

src/commonMain/kotlin/com/charleskorn/kaml/YamlInput.kt
EdwarDDay: To be able to decode a `YamlList` still as `YamlNode`. The `YamlNodeSerializer.descriptor` is polymorphic (because we don't know the type of it). So a list needs to be readable as polymorphic.

src/commonTest/kotlin/com/charleskorn/kaml/YamlWritingTest.kt
EdwarDDay: Oh, that was an error. Fixed it.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
```diff
@@ -0,0 +1,172 @@
...
+        value.content.singleOrNull()?.also { return encoder.encodeChar(it) }
+        encoder.encodeString(value.content)
```
charleskorn, line 83: Is there a test for this special handling of characters?

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
charleskorn: What if we just always encode the value as a string?

EdwarDDay: Then we would encode a double `1.2` as `"1.2"`, right? So the output might be unexpected to the user.

I could create `toBooleanOrNull`, `toByteOrNull`, …  functions in the `YamlScalar` class and use it in `toBoolean`, `toByte`, … functions and in the `YamlScalarSerializer`. This way the code doesn't get duplicated and we don't have to try catch everything.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
EdwarDDay: No, I just wanted to use the short version of `encodeChar`, to not have the overhead from `encodeString`. There are also no extra tests for `encodeBoolean`, `encodeLong`, `encodeDouble`. I thought, these are properly tested in the tests for `YamlOutput`.
But I can of course write extra tests, if you think, that's needed.

charleskorn: I think it'd be valuable to have a test that shows the expected behaviour when a value looks like it is a boolean or a number etc.

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
charleskorn: > I could create `toBooleanOrNull`, `toByteOrNull`, … functions in the `YamlScalar` class and use it in `toBoolean`, `toByte`, … functions and in the `YamlScalarSerializer`. This way the code doesn't get duplicated and we don't have to try catch everything.

This sounds good to me.

EdwarDDay: done

src/commonMain/kotlin/com/charleskorn/kaml/YamlNodeSerializer.kt
EdwarDDay: I added some tests for encoding the primitives via the new serializer.

charleskorn (approved):
Thanks @EdwarDDay!

PR #678, opened 2025-03-02
fix: don't throw ClassCastException when using scalar decoding methods with contextual descriptor

Stream29:
Delegating the behaviour of YamlContextualInput to YamlScalarInput when its node is YamlScalar and adding tests for it.

All tests passed.


src/commonMain/kotlin/com/charleskorn/kaml/YamlContextualInput.kt
```diff
@@ -23,16 +23,35 @@ import kotlinx.serialization.encoding.CompositeDecoder
 import kotlinx.serialization.modules.SerializersModule
 
 internal class YamlContextualInput(node: YamlNode, yaml: Yaml, context: SerializersModule, configuration: YamlConfiguration) : YamlInput(node, yaml, context, configuration) {
-    override fun decodeElementIndex(descriptor: SerialDescriptor): Int = throw IllegalStateException("Must call beginStructure() and use returned Decoder")
-    override fun decodeValue(): Any = when (node) {
```
charleskorn, line 27: Could you please add a test that exercises the previous behaviour of `decodeValue` and confirm that the changes here haven't broken this?

charleskorn (commented):
Thanks for the PR!

src/commonMain/kotlin/com/charleskorn/kaml/YamlContextualInput.kt
Stream29: I've already added tests for contextual deserialization for primitive types and enum types, which covers all the place `decodeValue` is used.

The usage of `decodeValue` is defined in `AbstractDecoder` and here's the source:

```
    public open fun decodeValue(): Any = throw SerializationException("${this::class} can't retrieve untyped values")

    override fun decodeNotNullMark(): Boolean = true
    override fun decodeNull(): Nothing? = null
    override fun decodeBoolean(): Boolean = decodeValue() as Boolean
    override fun decodeByte(): Byte = decodeValue() as Byte
    override fun decodeShort(): Short = decodeValue() as Short
    override fun decodeInt(): Int = decodeValue() as Int
    override fun decodeLong(): Long = decodeValue() as Long
    override fun decodeFloat(): Float = decodeValue() as Float
    override fun decodeDouble(): Double = decodeValue() as Double
    override fun decodeChar(): Char = decodeValue() as Char
    override fun decodeString(): String = decodeValue() as String
    override fun decodeEnum(enumDescriptor: SerialDescriptor): Int = decodeValue() as Int
```

In `src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt` all the usage above are tested correct.

So I think the tests has already covered all the needed usages.

charleskorn: This doesn't cover the case of custom serializers that relied on the old behaviour of `decodeValue()`.

Stream29: If my understanding is correct, there is no way that a custom serializer can invoke `AbstractDecoder.decodeValue(): Any` directly, unless it cast the given `Decoder` forcedly into `AbstractDecoder`.

The only legal way for custom serializer to depend on the behaviour of  `AbstractDecoder.decodeValue(): Any` is to call the methods delegated to `AbstractDecoder.decodeValue(): Any`, which ared listed as below:
```
    override fun decodeBoolean(): Boolean = decodeValue() as Boolean
    override fun decodeByte(): Byte = decodeValue() as Byte
    override fun decodeShort(): Short = decodeValue() as Short
    override fun decodeInt(): Int = decodeValue() as Int
    override fun decodeLong(): Long = decodeValue() as Long
    override fun decodeFloat(): Float = decodeValue() as Float
    override fun decodeDouble(): Double = decodeValue() as Double
    override fun decodeChar(): Char = decodeValue() as Char
    override fun decodeString(): String = decodeValue() as String
    override fun decodeEnum(enumDescriptor: SerialDescriptor): Int = decodeValue() as Int
```
In `YamlReadingTest.kt`, these usages are all verified.

Stream29: If I had made a mistake, you are welcomed to point out my mistake and I will accept that thankfully.

src/commonMain/kotlin/com/charleskorn/kaml/YamlContextualInput.kt
```diff
@@ -23,16 +23,35 @@ import kotlinx.serialization.encoding.CompositeDecoder
...
+
     override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder =
         createFor(node, yaml, serializersModule, configuration, descriptor)
 
     override fun getCurrentLocation(): Location = node.location
     override fun getCurrentPath(): YamlPath = node.path
+
+    private inline fun <T> delegateToYamlScalarInput(block: YamlScalarInput.() -> T): T {
+        delegateIfScalar?.let { return it.block() }
+        return when (node) {
```
charleskorn, line 52: Couldn't we inline `delegateIfScalar` here? This is the only place it's used.

Stream29: Agree. Let me do that.

src/commonMain/kotlin/com/charleskorn/kaml/YamlContextualInput.kt
```diff
@@ -22,17 +22,31 @@ import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.encoding.CompositeDecoder
 import kotlinx.serialization.modules.SerializersModule
 
-internal class YamlContextualInput(node: YamlNode, yaml: Yaml, context: SerializersModule, configuration: YamlConfiguration) : YamlInput(node, yaml, context, configuration) {
+internal class YamlContextualInput(node: YamlNode, yaml: Yaml, val context: SerializersModule, configuration: YamlConfiguration) : YamlInput(node, yaml, context, configuration) {
```
charleskorn, line 25: Why does this need to become a property? Can't we get the value of `context` from the `serializersModule` property on `YamlInput`?

```suggestion
internal class YamlContextualInput(node: YamlNode, yaml: Yaml, context: SerializersModule, configuration: YamlConfiguration) : YamlInput(node, yaml, context, configuration) {
```

Stream29: You are right. I'll be working on this.

charleskorn (commented):
Thanks @Stream29!

charleskorn (approved):

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -63,6 +63,7 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 import kotlinx.serialization.modules.serializersModuleOf
 import kotlin.jvm.JvmInline
+import kotlin.reflect.typeOf
```
charleskorn, line 66:
```suggestion
import kotlin.jvm.JvmInline
```

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -2386,6 +2406,22 @@ object ContextualSerializer : KSerializer<String> {
     override fun serialize(encoder: Encoder, value: String): Unit = throw UnsupportedOperationException()
 }
 
+class DelegatedContextualSerializer(private val serializer: KSerializer<Any>): KSerializer<Any> {
```
charleskorn, line 2409:
```suggestion
class DelegatedContextualSerializer(private val serializer: KSerializer<Any>) : KSerializer<Any> {
```

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -2386,6 +2406,22 @@ object ContextualSerializer : KSerializer<String> {
...
+    override val descriptor: SerialDescriptor = buildSerialDescriptor(
+        serialName = "DelegatedContextual<${serializer.descriptor.serialName}>",
+        kind = SerialKind.CONTEXTUAL
```
charleskorn, line 2412:
```suggestion
        kind = SerialKind.CONTEXTUAL,
```

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -2386,6 +2406,22 @@ object ContextualSerializer : KSerializer<String> {
...
+    )
+
+    override fun deserialize(decoder: Decoder): Any =
+        serializer.deserialize(decoder)
+
+
```
charleskorn, line 2418:
```suggestion

```

src/commonTest/kotlin/com/charleskorn/kaml/YamlReadingTest.kt
```diff
@@ -2386,6 +2406,22 @@ object ContextualSerializer : KSerializer<String> {
...
+    override fun serialize(encoder: Encoder, value: Any) {
+        serializer.serialize(encoder, value)
+    }
+
+}
```
charleskorn, line 2423:
```suggestion
    }
}
```

PR #721, opened 2025-06-28
feat: accept single value as list

WhiredPlanck:
(I connot run the test for "No JDK specified" exception, even I've specified, sorry)


WhiredPlanck:
I manage to run the tests, all yaml reading tests are passed.

图片

@charleskorn Would you like to have a look?

WhiredPlanck:
> Is there a reason why this needs to be implemented in Kaml rather than implemented as a custom serializer in your application?

My original intention to do this is to implement the same feature as Jackson. There are many "single values" I need to accept them as
 list in the data I want to deserialize. Some data are cloned from upstreams and I connot control the format. And If I want to deserialize a field as `Map<String, List<String>`, things seem to be complex to do with a custom serializer.

PR #777, opened 2025-10-21
feat: Preserve scalar styles and YAML tags when serializing YamlNode

### LINKED ISSUE #749: YamlNodeSerializer corrupts data due to implicit types conversion
### Describe the bug

Hi! YamlNode loses data and types due to implicit automatic conversion to numbers.

`"+44123456"` -> `44123456`

### Reproduction repo

N/A

### Steps to reproduce

```kotlin
@Test
fun YamlNode_reproducer() {
    val originalYaml = """|"user":
                          |    "phone_with_prefix": "+44123456"
                          |    "phone_without_prefix": "44123456"
                          |""".trimMargin()

    val yaml = Yaml(
        configuration = YamlConfiguration(
            singleLineStringStyle = SingleLineStringStyle.DoubleQuoted,
            ambiguousQuoteStyle = AmbiguousQuoteStyle.DoubleQuoted,
        )
    )
    val yamlNode = yaml.parseToYamlNode(originalYaml)

    val encodedYaml = yaml.encodeToString(yamlNode)

    //  "phone_with_prefix": 44123456    <-- "+" is lost
    //  "phone_without_prefix": 44123456 <-- string type is lost
    expect(encodedYaml).toEqual(originalYaml)
}
```

### Expected behaviour

Quoted strings kept as is, not losing content.

### Actual behaviour

1. Type is lost (quoted string becomes a number).
2. String's content is modified

### Version information

```plain text
0.85.0
```

### Any other information

_No response_

Tomas-Jakubec:
### Summary  
Fixes loss of scalar style and YAML tag information during round\-trip serialization of `YamlNode`s. Resolves issue https://github.com/charleskorn/kaml/issues/749.

### Problem  
Currently `YamlScalar` does not retain the original scalar style, making it impossible to distinguish between:
- A plain numeric scalar (e.g. 123)  
- A quoted string that looks like a number (e.g. "123")  
Additionally, YAML tags present in the parsed input were not emitted during serialization.

### Root cause  
- Scalar style (plain, single quoted, double quoted, literal, folded, JSON style) was not stored alongside the scalar value.  
- Serialization of `YamlNode` relied on generic Kotlinx Serialization paths that discarded style and tag metadata.

### Solution  
- Introduced preservation of scalar style via an explicit enum (`YamlNodeScalarStyle`) mapped to SnakeYAML engine styles.  
- Ensured YAML tags on nodes are carried through and emitted.  
- Delegated encoding of `YamlNode` subclasses to `YamlOutput`, separating it from standard Kotlinx Serialization logic to avoid interference and to centralize style/tag handling.

### Key changes   
- Updated serialization pipeline to include style and tag emission.  
- Adjusted `YamlOutput` to take responsibility for `YamlNode` encoding.  
- Added/updated tests to assert round\-trip fidelity for:  
  - Quoted vs plain numeric\-looking scalars  
  - Block scalar styles (literal, folded)  
  - Explicit tags  

Backward compatibility  
- Public API surface minimally affected (new enum exposed).  
- Existing consumers not relying on style/tag round\-tripping remain unaffected.  
- No breaking rename/removal of existing types.

Performance  
- Negligible impact; added metadata handling only during node serialization.

Risks / considerations  
- Consumers constructing `YamlScalar` programmatically may now wish to specify a style explicitly.  However, not specifying it results in original behavior.
- Ensure all new styles are covered in multiplatform targets. I tested all platforms except for tvosSimulatorArm64Test, please check.

Testing  
- Added unit tests for style preservation and tag emission.  
- Verified that previously failing scenario from issue \#749 now passes.

Closes: \#749

Let me know if a changelog entry format needs adjustment.

### DIFF
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
@@ -18,6 +18,7 @@
 
 package com.charleskorn.kaml
 
+import it.krzeminski.snakeyaml.engine.kmp.common.ScalarStyle
 import kotlinx.serialization.Serializable
 
 @Serializable(with = YamlNodeSerializer::class)
@@ -34,7 +35,7 @@ public sealed class YamlNode(public open val path: YamlPath) {
 }
 
 @Serializable(with = YamlScalarSerializer::class)
-public data class YamlScalar(val content: String, override val path: YamlPath) : YamlNode(path) {
+public data class YamlScalar(val content: String, override val path: YamlPath, val scalarStyle: YamlNodeScalarStyle? = null) : YamlNode(path) {
     override fun equivalentContentTo(other: YamlNode): Boolean = other is YamlScalar && this.content == other.content
     override fun contentToString(): String = "'$content'"
 
--- a/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
+++ b/src/commonMain/kotlin/com/charleskorn/kaml/YamlNode.kt
@@ -18,7 +18,6 @@
 
 package com.charleskorn.kaml
 
-import it.krzeminski.snakeyaml.engine.kmp.common.ScalarStyle
 import kotlinx.serialization.Serializable
 
 @Serializable(with = YamlNodeSerializer::class)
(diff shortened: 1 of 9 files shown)


Tomas-Jakubec:
@charleskorn

